PR #43, opened 2020-04-21
Fix concurrency issue with TimeSeriesUploadQueue

mathialo:
If a new time series is added during the list comprehension in upload()
the iterator for the dict view will fail. Move everything inside lock.


mathialo:
> Use `with self.lock`?

Yes, I'm planning to go through the entire lib and do that throughout.

PR #47, opened 2020-05-15
Several minor additions to configtools

mathialo:
- Add a separator bewteen thread name and message in log lines
 - Add defaults to StateStoreConfig
 - Add optional data set id to cognite section
 - Add int as an option to version, and make optional
 - create_state_store never returns None, add NoStateStore

### DIFF
--- a/cognite/extractorutils/configtools.py
+++ b/cognite/extractorutils/configtools.py
@@ -358,8 +358,8 @@ class LocalStateStoreConfig:
 
 @dataclass
 class StateStoreConfig:
-    raw: Optional[RawStateStoreConfig]
-    local: Optional[LocalStateStoreConfig]
+    raw: Optional[RawStateStoreConfig] = None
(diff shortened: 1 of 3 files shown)


cognite/extractorutils/configtools.py
```diff
@@ -163,6 +163,7 @@ class CogniteConfig:
     project: str
     api_key: Optional[str]
     idp_authentication: Optional[AuthenticatorConfig]
+    data_set_id: Optional[int]
```
lprovensi, line 166: For the dotnet utils, we decided to leave` data-set-id` and `external-id-prefix` out of the lib because we were not enforcing these on time series creation. That is, the client applications set those properties. Not sure if it is the same for python

mathialo: Yeah, I'm not really sold on it being there for all extractors either, but for the `BaseConfig` class to make sense, I believe this is the best approach for now.

PR #59, opened 2020-07-27
Add long living CDF client

pavelzubarev:


pavelzubarev:
That doesn't really work.

PR #63, opened 2020-07-29
Remove shadow of stdlib typing in Python > 3.5

alexchandel:
The `typing` package is builtin from 3.5 on.  Installing it as a dependency causes all sorts of weird problems & errors, for example [all these](https://github.com/python/typing/issues/573).  Only depend on it for Python < 3.5.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -437,11 +437,11 @@ description = "Powerful data structures for data analysis, time series, and stat
 name = "pandas"
 optional = false
 python-versions = ">=3.6.1"
-version = "1.0.5"
+version = "1.1.0"
 
 [package.dependencies]
-numpy = ">=1.13.3"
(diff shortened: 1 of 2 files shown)


mathialo:
Thanks for the PR, nice catch!

Since we're requiring at least `3.6` of Python itself, I guess we don't even need the dependency at all. Could you just remove the line alltogether? Also remember to run `poetry lock` as we're dependent on the lock file to install dependencies in the GitHub Actions pipeline.

Btw, if you are wondering why the builds fail, it's because external PRs lack some credentials necessary for integration tests, don't worry about it 🙂

alexchandel:
That's good, those build errors looked pretty incredible for a no-op type-hint change.  I force-pushed an amended commit with the dep removed & lockfile updated.

PR #66, opened 2020-09-03
Added sequence uploader

thomafred:


cognite/extractorutils/uploader.py
```diff
@@ -560,3 +560,264 @@ def __len__(self) -> int:
...
+
+    @retry(
+        exceptions=CogniteAPIError,
+        tries=RETRIES,
+        delay=RETRY_DELAY,
+        max_delay=RETRY_MAX_DELAY,
+        backoff=RETRY_BACKOFF_FACTOR,
+    )
+    def _upload_single(self, either_id: EitherId, upload_this: SequenceData) -> SequenceData:
+        self._create_or_update(either_id)
```
mathialo, line 738: Calling this every time will double the amount of API requests, slowing down the extractor.

The pattern used elsewhere is to call the insert method, and create the sequence on an error. This reduces the amount of requests with quite a bit.

cognite/extractorutils/uploader.py
```diff
@@ -560,3 +560,264 @@ def __len__(self) -> int:
             Number of events in queue
         """
         return self.upload_queue_size
+
+
+class SequenceUploadQueue(AbstractUploadQueue):
+
+    def __init__(
```
mathialo, line 567: The `TimeSeriesUploadQueue` has a `create_missing` option to enable the automatic creation of missing time series. For consistency, I suggest adding that as an option here as well, and leave it off by default.

mathialo (changes requested):
Great addition! 

See comments for a couple of thoughts on consistency with the rest of the library. Would also be great with either a unit or an integration test.

It also seems that `black` is complaining about the format, a simple `poetry run black` should fix that :)

thomafred:
I have made the requested changes, however the TimeSeries integration tests are failing due to missing API key. I do not have access to an environment where I can run these tests.

mathialo:
> I have made the requested changes, however the TimeSeries integration tests are failing due to missing API key. I do not have access to an environment where I can run these tests.

Nice! Thanks a lot for the contribution.

No worries, that's a known issue for PRs on this repo. The tests will pass on master 🙂

PR #67, opened 2020-09-06
Created file upload queue

thomafred:
Added uplpad-queue for the "File"-resource type.

Unit-test included this time :)

### DIFF
--- a/cognite/extractorutils/uploader.py
+++ b/cognite/extractorutils/uploader.py
@@ -47,9 +47,9 @@ from requests.exceptions import ConnectionError
 from retry import retry
 
 from cognite.client import CogniteClient
-from cognite.client.data_classes import Event, TimeSeries, Sequence, SequenceData
+from cognite.client.data_classes import Event, Sequence, SequenceData, TimeSeries
 from cognite.client.data_classes.raw import Row
-from cognite.client.exceptions import CogniteAPIError, CogniteNotFoundError, CogniteDuplicatedError
+from cognite.client.exceptions import CogniteAPIError, CogniteDuplicatedError, CogniteNotFoundError
 
 from ._inner_util import _resolve_log_level
 from .util import EitherId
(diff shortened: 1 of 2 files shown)


cognite/extractorutils/uploader.py
```diff
@@ -809,3 +810,115 @@ def __len__(self) -> int:
...
+        backoff=RETRY_BACKOFF_FACTOR,
+    )
+    def _upload_batch(self):
+        for i, (file_meta, file_name) in enumerate(self.upload_queue):
+
+            # Upload file
+            file_meta = self.cdf_client.files.upload(file_name, **file_meta.dump(), overwrite=self.overwrite_existing)
+
+            # Update meta-object in queue
+            self.upload_queue[i] = (file_meta, file_name)
```
mathialo, line 894: Why update this when we're clearing the queue right after?

cognite/extractorutils/uploader.py
```diff
@@ -809,3 +810,115 @@ def __len__(self) -> int:
...
+        tries=RETRIES,
+        delay=RETRY_DELAY,
+        max_delay=RETRY_MAX_DELAY,
+        backoff=RETRY_BACKOFF_FACTOR,
+    )
+    def _upload_batch(self):
+        for i, (file_meta, file_name) in enumerate(self.upload_queue):
+
+            # Upload file
+            file_meta = self.cdf_client.files.upload(file_name, **file_meta.dump(), overwrite=self.overwrite_existing)
```
mathialo, line 891: For all other resource types, the SDK is doing some chunking and parallelization of requests, but since the API doesn't support multiple files in a single upload, it doesn't do it for files. 

I think it would make sense for us to add some parallelization to this since we're batching up files for uploading, something like
```python
with ThreadPoolExecutor(self.cdf_client.config.max_workers) as pool:
    for i, (file_meta, file_name) in enumerate(self.upload_queue):
        pool.submit(fn=self._upload_single, args=(file_name, file_meta))
```
Maybe also move the retry logic to the individual uploads to avoid redoing all the uploading on a single failure.

cognite/extractorutils/uploader.py
thomafred: The post-upload function is called before the queue is cleared

mathialo: Good point

cognite/extractorutils/uploader.py
thomafred: This is a good idea. I have updated this function to reflect this.

mathialo (approved):
Very nice!

Issue #69, opened 2020-09-08
RawUploadQueue not handling CogniteReadTimeout


thomafred:
The `RawUploadQueue` does not handle `CogniteReadTimeout`-exception, sometimes thrown by `RawRowsAPI.insert`.

Trace (application-specific frames omitted):

```python
---------------------------------------------------------------------------
timeout                                   Traceback (most recent call last)
/databricks/python/lib/python3.7/site-packages/urllib3/connectionpool.py in _make_request(self, conn, method, url, timeout, chunked, **httplib_request_kw)
    420                     # Otherwise it looks like a bug in the code.
--> 421                     six.raise_from(e, None)
    422         except (SocketTimeout, BaseSSLError, SocketError) as e:

/databricks/python/lib/python3.7/site-packages/urllib3/packages/six.py in raise_from(value, from_value)

/databricks/python/lib/python3.7/site-packages/urllib3/connectionpool.py in _make_request(self, conn, method, url, timeout, chunked, **httplib_request_kw)
    415                 try:
--> 416                     httplib_response = conn.getresponse()
    417                 except BaseException as e:

/usr/lib/python3.7/http/client.py in getresponse(self)
   1343             try:
-> 1344                 response.begin()
   1345             except ConnectionError:

/usr/lib/python3.7/http/client.py in begin(self)
    305         while True:
--> 306             version, status, reason = self._read_status()
    307             if status != CONTINUE:

/usr/lib/python3.7/http/client.py in _read_status(self)
    266     def _read_status(self):
--> 267         line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
    268         if len(line) > _MAXLINE:

/usr/lib/python3.7/socket.py in readinto(self, b)
    588             try:
--> 589                 return self._sock.recv_into(b)
    590             except timeout:

/databricks/python/lib/python3.7/site-packages/urllib3/contrib/pyopenssl.py in recv_into(self, *args, **kwargs)
    325             if not util.wait_for_read(self.socket, self.socket.gettimeout()):
--> 326                 raise timeout("The read operation timed out")
    327             else:

timeout: The read operation timed out

During handling of the above exception, another exception occurred:

ReadTimeoutError                          Traceback (most recent call last)
/databricks/python/lib/python3.7/site-packages/requests/adapters.py in send(self, request, stream, timeout, verify, cert, proxies)
    448                     retries=self.max_retries,
--> 449                     timeout=timeout
    450                 )

[...]

PR #70, opened 2020-09-08
Addressing #69 - Raw upload queue cognite read timeout

thomafred:
PR addressing #69 

Also fixed failing unit-test.


mathialo (approved):
Nice catch!

mathialo (changes requested):
It seems the CogniteReadTimeout exception was added recently (v2.2.2). The import brakes with any version of the SDK before that.

Could you change the dependency in `pyproject.toml` from `cognite-sdk = ">=1.4.8<3.0.0"` to `cognite-sdk = "^2.2.2"` and run `poetry lock`?

PR #71, opened 2020-09-09
Release version 1.2.0

mathialo:
* Add upload queues for sequences and files
 * TimeSeriesUploadQueue can now auto-create string time series

### DIFF
--- a/cognite/extractorutils/__init__.py
+++ b/cognite/extractorutils/__init__.py
@@ -16,4 +16,4 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "1.0.3"
+__version__ = "1.2.0"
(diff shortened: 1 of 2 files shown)


cognite/extractorutils/__init__.py
```diff
@@ -16,4 +16,4 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "1.0.3"
+__version__ = "1.2.0"
```
sebbert, line 19: Would it be possible to read this value from `pyproject.toml` to avoid having two divergent sources of truth here?

sebbert (approved):

cognite/extractorutils/__init__.py
mathialo: I know, I really wish poetry somehow could populate this automatically, however the poetry folks have been [quite clear that they're never doing that](https://github.com/python-poetry/poetry/issues/144), and I have not found any way to work around this that I find sufficiently elegant.

PR #76, opened 2020-12-01
Add general OAuth2/OIDC option

mathialo:
Add option to provide a token url directly, instaed of an AAD tenant.
Also provide an (optional) resource parameter.


hakontro (approved):
LGTM!

PR #78, opened 2021-01-05
Fix pool issue in FileUploadQueue

mathialo:
FileUploadQueue had an incorrect call to the threadpool's submit that
caused the files to never be uploaded.

### DIFF
--- a/cognite/extractorutils/uploader.py
+++ b/cognite/extractorutils/uploader.py
@@ -893,21 +893,18 @@ class FileUploadQueue(AbstractUploadQueue):
         backoff=RETRY_BACKOFF_FACTOR,
     )
     def _upload_single(self, index, file_name, file_meta):
-
         # Upload file
-        file_meta = self.cdf_client.files.upload(file_name, **file_meta.dump(), overwrite=self.overwrite_existing)
+        file_meta = self.cdf_client.files.upload(file_name, overwrite=self.overwrite_existing, **file_meta.dump())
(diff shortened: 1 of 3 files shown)


cognite/extractorutils/uploader.py
```diff
@@ -893,21 +893,18 @@ def upload(self) -> None:
         backoff=RETRY_BACKOFF_FACTOR,
     )
     def _upload_single(self, index, file_name, file_meta):
-
         # Upload file
-        file_meta = self.cdf_client.files.upload(file_name, **file_meta.dump(), overwrite=self.overwrite_existing)
+        file_meta = self.cdf_client.files.upload(file_name, overwrite=self.overwrite_existing, **file_meta.dump())
```
emelvarol-cognite, line 897: I might be wrong, is it ok to use keyword arguments for all? I thought it causes error to use positional after keyword arguments.

mathialo: the `**` syntax expands a dictionary to be passed as keyword arguments, so
``` python
foo(a=1, b=2)
```
and
``` python
d = {'a': 1, 'b': 2}
foo(**d)
```
is the same, so the last argument is really a large set of keyword arguments in disguise :)

emelvarol-cognite: Ehe got it :D

emelvarol-cognite (approved):

PR #82, opened 2021-03-01
Fix URL creation in OAuth flow

mathialo:
It was appending `/oauth2/v2.0/token` always, leading to it being
appended twice when using AAD tenants, and incorrectly appended to urls
when specifying token url directly.


einarmo (approved):
LGTM

PR #84, opened 2021-03-12
Add a single cancelation token + enhance metrics for upload queue

pavelzubarev:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cognite-extractor-utils"
-version = "1.2.4"
+version = "1.2.5"
```
mathialo, line 3: This should at least be a minor version change IMO

cognite/extractorutils/uploader.py
```diff
@@ -175,11 +182,86 @@ def stop(self, ensure_upload: bool = True) -> None:
...
             self.upload()
 
 
+@dataclass(frozen=True)
+class TimestampedObject:
+    payload: Any
+    created: Arrow
+
+
+raw_uploader_rows_queued = Counter(
```
mathialo, line 196: Should these be capitalized (ie, do we consider these constants)? My gut says so since we don't expect users (or ourselves) to overwrite the contents of these variables.

cognite/extractorutils/uploader.py
```diff
@@ -203,12 +285,29 @@ def __init__(
...
+            cancelation_token,
         )
-        self.upload_queue: Dict[str, Dict[str, List[Row]]] = dict()
+        self.upload_queue: Dict[str, Dict[str, List[TimestampedObject]]] = dict()
+
+        # It is a hack since Prometheus client registers metrics on object creation, so object has to be created once
+        # Store the reference to a global objects.
+        global raw_uploader_rows_queued, raw_uploader_rows_written, raw_uploader_rows_duplicates
+        global raw_uploader_queue_size, raw_uploader_latency
+        self.rows_queued = raw_uploader_rows_queued
```
mathialo, line 306: Why add these as instance attributes, and not just use the global variables directly in the other methods?

mathialo (commented):
A lot of really nice changes here! Just a few comments.

cognite/extractorutils/uploader.py
pavelzubarev: I was thinking to make it clear that those come outside, but strictly speaking, we can rename and drop the global declaration

mathialo: I was thinking more of why we add `self.rows_queued` instead of just using `raw_uploader_rows_queued` directly

mathialo: If you think it's cleaner, that's fine 🙂 I was just wondering what the reasoning was.

mathialo: Nevermind, seeing the new changes after the renaming changed my mind 🙂 let's keep them

mathialo (approved):
Great!

PR #87, opened 2021-03-29
Reduce metrics cardinality

pavelzubarev:
It was a bad idea to have timeseries external_id as a label.


lprovensi (approved):
I can see this was a bad idea. Hopefully there are no boards relying on these labels

pavelzubarev:
> I can see this was a bad idea. Hopefully there are no boards relying on these labels

there is no.. it's quite new change

PR #98, opened 2021-09-20
BytesUploadQueue

kristogj:
I know this is very similar to FilesUploadQueue, but this is very useful when we are building extractors that make requests to surveillance cameras for example.

### DIFF
--- a/cognite/extractorutils/uploader.py
+++ b/cognite/extractorutils/uploader.py
@@ -257,6 +257,13 @@ FILES_UPLOADER_LATENCY = Histogram(
     "cognite_files_uploader_latency", "Distribution of times in minutes records spend in the queue",
 )
 
+BYTES_UPLOADER_QUEUED = Counter("cognite_bytes_uploader_queued", "Total number of frames queued")
+BYTES_UPLOADER_WRITTEN = Counter("cognite_bytes_uploader_written", "Total number of frames written")
+BYTES_UPLOADER_QUEUE_SIZE = Gauge("cognite_bytes_uploader_queue_size", "Internal queue size")
+BYTES_UPLOADER_LATENCY = Histogram(
+    "cognite_bytes_uploader_latency", "Distribution of times in minutes records spend in the queue",
+)
+
 
 class RawUploadQueue(AbstractUploadQueue):
     """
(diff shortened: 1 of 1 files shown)


cognite/extractorutils/uploader.py
```diff
@@ -1166,3 +1173,134 @@ def __len__(self) -> int:
             Number of events in queue
         """
         return self.upload_queue_size
+
+
+class BytesUploadQueue(AbstractUploadQueue):
```
mathialo, line 1178: Please add a short docstring for the class as well

cognite/extractorutils/uploader.py
```diff
@@ -1166,3 +1173,134 @@ def __len__(self) -> int:
...
+        self.queue_size = BYTES_UPLOADER_QUEUE_SIZE
+        self.latency = BYTES_UPLOADER_LATENCY
+        self.bytes_written = BYTES_UPLOADER_WRITTEN
+
+    def add_to_upload_queue(self, frame: bytes, metadata: FileMetadata) -> None:
+        """
+        Add the frame to the upload queue
+        :param frame: bytes
+        :param metadata:
+        :return:
```
mathialo, line 1215: This project uses Google-style docstrings, which should look like this:
```
Add a frame to the upload queue

Args:
    frame: <short description>
    metadata: <short description>
```

cognite/extractorutils/uploader.py
```diff
@@ -1166,3 +1173,134 @@ def __len__(self) -> int:
...
+        with self.lock:
+            if self.upload_queue_size == 0:
+                self.latency_zero_point = arrow.utcnow()
+
+            self.upload_queue.append((frame, metadata))
+            self.upload_queue_size += 1
+            self.bytes_queued.inc()
+            self.queue_size.set(self.upload_queue_size)
+
+    def upload(self) -> None:
```
mathialo, line 1226: I wonder if we could generalize and reuse some parts of the fileuploader to avoid duplicating code for the upload part, but I suggest we do a separate refactoring of that later

mathialo (commented):
Thanks for your contribution!

Please also add a test for the new queue (create another test case in [tests/tests_unit/test_cdf_upload_queues.py](https://github.com/cognitedata/python-extractor-utils/blob/master/tests/tests_unit/test_cdf_upload_queues.py) similar to those already there)

cognite/extractorutils/uploader.py
kristogj: I was thinking about that same, but agree, we can refactor that later 💪🏻

PR #99, opened 2021-09-21
Add throttled_loop function

mathialo:
This will allow for main loops to easily be configured to run with a
target iteration time, so that we don't overload the source system, or
don't request the same info many times (for source systems that only
have new data every x seconds).

Example:
``` python
for _ in throttled_loop(10, Event()):
    print("Hello every 10 seconds!")
```


einarmo (approved):
Nice. Async "yield" is actually a pretty neat way to do this.

PR #100, opened 2021-09-22
delete login and use config to get project

tgib23:
Using login will cause error when accessing with OIDC configuration.

https://cognitedata.atlassian.net/browse/CDF-13268

Because this change doesn't seem to affect current tests, I didn't add any test code.
If necessary, I'll do.


mathialo (approved):
Thanks! Good catch 🙂

PR #103, opened 2021-09-24
Add option to resolve data set either from id or external id

mathialo:
Many SAs have asked for an option to use external ID in config files
as it stays the same accross CDF projects. That will for sure be the
case for extraction pipelines too, so introduce a new more generic
schema for data sets, that we can reuse for extraction pipelines later


einarmo (approved):
LGTM!

PR #104, opened 2021-09-24
Release 1.5.1

mathialo:


einarmo (approved):
Have you ever considered using git tags to do this?

mathialo:
> Have you ever considered using git tags to do this?

Thing is, we would need to update these as well because of packaging and publishing. But I have definitely been thinking we should tag _as well_ so the releases page is automatically updated. Right now it's quite out of sync again

PR #114, opened 2021-11-24
Methods to remove invalid datapoints

einarmo:
Also metrics, and a test to verify that this works.

The rules are:

No NaN or +/-Inf

Points in range (-1E100, 1E100)

String datapoints no longer than 255 characters.

Timestamps after 01.01.1971

### DIFF
--- a/cognite/extractorutils/uploader.py
+++ b/cognite/extractorutils/uploader.py
@@ -65,6 +65,7 @@ instead. If both are used, the condition being met first will trigger the upload
 """
 
 import logging
+import math
 import threading
 import time
 from abc import ABC, abstractmethod
@@ -249,6 +250,11 @@ TIMESERIES_UPLOADER_QUEUE_SIZE = Gauge("cognite_timeseries_uploader_queue_size",
 
 TIMESERIES_UPLOADER_LATENCY = Histogram(
     "cognite_timeseries_uploader_latency", "Distribution of times in minutes records spend in the queue",
+)
+
+TIMESERIES_UPLOADER_POINTS_DISCARDED = Counter(
+    "cognite_timeseries_uploader_points_discarded",
+    "Total number of datapoints discarded due to invalid timestamp or value",
 )
 
 SEQUENCES_UPLOADER_POINTS_QUEUED = Counter(
@@ -456,6 +462,12 @@ def default_time_series_factory(external_id: str, datapoints: DataPointList) -> 
     return TimeSeries(external_id=external_id, is_string=is_string)
 
 
+MIN_DATAPOINT_TIMESTAMP = 31536000000
+MAX_DATAPOINT_STRING_LENGTH = 255
+MAX_DATAPOINT_VALUE = 1e100
+MIN_DATAPOINT_VALUE = -1e100
+
+
 class TimeSeriesUploadQueue(AbstractUploadQueue):
     """
     Upload queue for time series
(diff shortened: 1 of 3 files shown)


cognite/extractorutils/uploader.py
```diff
@@ -512,6 +524,37 @@ def __init__(
...
+            Dict[Union[int, float, datetime], Union[int, float, str]],
+            Tuple[Union[int, float, datetime], Union[int, float, str]],
+        ],
+    ) -> bool:
+        if isinstance(dp, Dict):
+            return self._verify_datapoint_time(dp["timestamp"]) and self._verify_datapoint_value(dp["value"])
+        elif isinstance(dp, Tuple):
+            return self._verify_datapoint_time(dp[0]) and self._verify_datapoint_value(dp[1])
+        else:
+            # I don't know how many different legal permutations there are, since the type hints are incomplete.
```
mathialo, line 555: I'm a bit unsure what you mean by incomplete type hints? If the user sends arguments that doesn't match the type hints, that's on them, and I think just returning here (like we are) is fine.

einarmo: You are right, of course. I wrote this with a slight misunderstanding, the type hints for DataPointList are actually wrong, from the SDK.

cognite/extractorutils/uploader.py
```diff
@@ -512,6 +524,37 @@ def __init__(
...
+                math.isnan(value) or math.isinf(value) or value > MAX_DATAPOINT_VALUE or value < MIN_DATAPOINT_VALUE
+            )
+        elif isinstance(value, str):
+            return len(value) <= MAX_DATAPOINT_STRING_LENGTH
+        else:
+            return True
+
+    def _is_datapoint_valid(
+        self,
+        dp: Union[
```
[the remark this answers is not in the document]
einarmo: There isn't an alias for the contents of DataPointList. Is it defined here or in the SDK? If it's from the utils I'll do something about this.

einarmo:
I fixed the type of DataPointList. I think what is there now is correct, but you should probably double check. It is slightly different in that it is no longer a `Union[List, List]`, but a `List[Union]`, which is more general and technically more correct, I think?

cognite/extractorutils/uploader.py
```diff
@@ -93,10 +94,10 @@
 RETRY_DELAY = 5
 RETRIES = 10
 
-DataPointList = Union[
-    List[Dict[Union[int, float, datetime], Union[int, float, str]]],
-    List[Tuple[Union[int, float, datetime], Union[int, float, str]]],
+DataPoint = Union[
+    Dict[str, Union[int, float, str, datetime]], Tuple[Union[int, float, datetime], Union[int, float, str]]
 ]
+DataPointList = List[DataPoint]
```
mathialo, line 100: I believe this is too generic, it would allow e.g.

```
[(ts1, val1), {ts2: val2}]
```

which I don't think the SDK allows?

mathialo: This is more generic though, so were not breaking anything, but I believe there are things that would type check and then break when run - but I actually think that might already be the case for the previous type hint.

Maybe we should just make it much simpler. Just doing a list of tuples or something like that, and de-generify it a little bit. That might be nice for readability and keeping it understandable. That's a breaking change though, so maybe merge this as-is, and then do a redesign of this whole typing situation for a 2.0.0 release.

mathialo (approved):

cognite/extractorutils/uploader.py
einarmo: Urgh. You're right. It's a `List[Union]` vs `Union[List]` thing again... I guess we'll merge it as is.

PR #115, opened 2021-11-24
Add periodic calls to synchronize for state-store

einarmo:
Uses the existing config option, and the same syntax as upload queues.

### DIFF
--- a/cognite/extractorutils/configtools.py
+++ b/cognite/extractorutils/configtools.py
@@ -429,7 +429,7 @@ class RawDestinationConfig:
 
 @dataclass
 class RawStateStoreConfig(RawDestinationConfig):
-    upload_interval: int = 30
+    save_interval: int = 30
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


cognite/extractorutils/configtools.py
```diff
@@ -429,7 +429,7 @@ class RawDestinationConfig:
 
 @dataclass
 class RawStateStoreConfig(RawDestinationConfig):
-    upload_interval: int = 30
+    save_interval: int = 30
```
mathialo, line 432: I get that it's unfortunate that the naming is a bit inconsistent between the two state stores, but this is breaking. An extractor with the `upload-interval` field configured would be blocked from starting after this change.

What we could do is introduce a new `save-interval` field, update all our docs to use it, and add a deprecation warning whenever someone is using the old `upload-interval`.

einarmo: Fair. I figured since it wasn't used it was fine, but I'll do as you suggest, and use either.

einarmo: Ugh. We can't even do that properly, because it has a default value of 30, so we would have to check if save_interval is the default value before using it instead. I think we best just keep upload_interval.

mathialo: Let's add it to the list of changes we want to do for 2.0!

mathialo (approved):
LGTM

PR #117, opened 2022-01-19
Make sure bool conversion is performed after env var expansion

emelvarol-cognite:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cognite-extractor-utils"
-version = "1.5.4"
+version = "1.5.5"
```
mathialo, line 3: There's a __version__ variable in the __init__ file that also should be updated :) 

Really wish poetry could do that automatically..

emelvarol-cognite: this PR will not work for us anyway, casting bool is not safe as bool("false")= True or bool("text")=True :/ Did some search on how we can customize the casting here but couldn't find anything.

cognite/extractorutils/configtools.py
```diff
@@ -503,9 +504,14 @@ def load_yaml(
...
     """
 
-    def env_constructor(_, node):
-        # Expnadvars uses same syntax as our env var substitution
-        return os.path.expandvars(node.value)
+    def env_constructor(l: yaml.SafeLoader, node):
+        expanded_value = os.path.expandvars(node.value)
+        new_node = copy.deepcopy(node)
+        new_node.value = expanded_value
+        return l.construct_yaml_bool(new_node)
```
emelvarol-cognite, line 511: It coult be just fine to do this maybe 🤔 Wasn't sure if we want to modify the original node but deepcopy is also not really nice. I don't like this solution as well because it will only work for string and bool environment variable conversion but couldn't find a way to enforce our constructor to run first.  @mathialo 
```suggestion
    def env_constructor(l: yaml.SafeLoader, node):
        expanded_value = os.path.expandvars(node.value)
        node.value = expanded_value
        return l.construct_yaml_bool(new_node)
```

cognite/extractorutils/configtools.py
```diff
@@ -503,9 +503,20 @@ def load_yaml(
...
+        bool_values = {
+            "yes": True,
+            "no": False,
+            "true": True,
+            "false": False,
+            "on": True,
+            "off": False,
+        }
+        expanded_value = os.path.expandvars(node.value)
+        return bool_values.get(expanded_value.lower(), node.value)
```
emelvarol-cognite, line 516: Honestly, this is not the ideal solution. We want to replace env vars first and then call default `yaml `constructors for type inference.  But haven't seen a way to enforce our `env_contructor` to run first so this is just a solution for `env var str -> bool`. Numeric conversions, and the others still will not work. 

Maybe we should manipulate `source` and find/replace env var syntaxes but idk.

cognite/extractorutils/configtools.py
```diff
@@ -503,9 +503,20 @@ def load_yaml(
...
-        # Expnadvars uses same syntax as our env var substitution
-        return os.path.expandvars(node.value)
+    def env_constructor(_: yaml.SafeLoader, node):
+        bool_values = {
+            "yes": True,
+            "no": False,
+            "true": True,
+            "false": False,
+            "on": True,
+            "off": False,
```
mathialo, line 513: While I don't see anyone requiring the string `"true"`, I'm not sure always parsing e.g. `"on"` as `True` is safe

I think sticking with true/false is best

emelvarol-cognite: hehe took the list from pyyaml bool constructor, I can narrow it down 👍

pyproject.toml
mathialo, line 3: There are some other significant changes on master already, so I suggest getting #114 through, and release 1.6.0 🙂

emelvarol-cognite: ok, I'll wait for it then. So that PR will not do version bump, right?

emelvarol-cognite: I'll make it 1.6.0 here when #114  goes into master

mathialo (approved):
🚀

PR #122, opened 2022-01-24
Add option to not handle interrupts

mathialo:
The win32 library for building windows services runs everything in
dummy threads, so setting the signal handler (which must be done from
the main thread) can't be done.

Add both an option to skip it (for services), and add a try/catch to
avoid the issue of a wrong signal handler crashing the extractor in the
future.


einarmo (approved):
Fun. LGTM.

PR #125, opened 2022-01-25
Added ingestion of .env

CogniteJohan:
If a .env file exists in working directory, it will upload the environment variables and possibly update existing ones with new values. This also gives a log message that no file was found to make environment variables, which is useful since without these one gets a cryptic error message when trying to make a Cognite Client and get confusing messages about api-keys etc.


CogniteJohan:
I did not figure out a good place to document that one could have a .env file in the working directory. Maybe in extractor utils documentation?

CogniteJohan:
@mathialo , not sure how to best resolve the conflict in poetry.lock. One hash is different. I thought I had to update it in order for the build tests to create correct environment.

PR #126, opened 2022-02-01
Use threading tokens in retry

einarmo:
This fixes the issue that retries would block the entire application due to GIL madness or something along those lines.

The solution isn't ideal. I couldn't find a great way to pass cancelation_token into the decorator, so now it just creates a new one. Suggestions are welcome.

### DIFF
--- a/cognite/extractorutils/retry.py
+++ b/cognite/extractorutils/retry.py
@@ -58,17 +58,17 @@ def retry(
 
     This is adapted from https://github.com/invl/retry
 
-    :param cancelation_token: a threading token that is waited on.
-    :param exceptions: an exception or a tuple of exceptions to catch. default: Exception.
-    :param tries: the maximum number of attempts. default: -1 (infinite).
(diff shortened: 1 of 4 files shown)


cognite/extractorutils/retry.py
```diff
@@ -0,0 +1,83 @@
...
+    max_delay: Optional[int] = None,
+    backoff: float = 1,
+    jitter: float = 0,
+    logger: logging.Logger = logging_logger,
+):
+    """Returns a retry decorator.
+
+    This is adapted from https://github.com/invl/retry
+
+    :param cancelation_token: a threading token that is waited on.
```
mathialo, line 61: We use google-style docstrings. Probably should have that in a contribution note somewhere

einarmo: I'll migrate them. This is straight up taken from the original repo.

mathialo (approved):
Nice!

mathialo:
Let's merge this in before #127 to include this in the release

PR #130, opened 2022-02-08
Always load LocalStateStore when requested

mathialo:
From the docs:

> If found, it will use that configuration to generate a state store,
> if no such config is found it will either create a LocalStateStore or
> a NoStateStore depending on whether the  `use_default_state_store`
> argument to the constructor was true or false.

If no `StateStoreConfig` was found, then there would be no
`LocalStateStore` created, contradicting the docs.

This should make the `LocalStateStore` always initialize/load if the
`use_default_state_store` argument was set to `True` and only load a
`NoStateStore` if set to `False`.


einarmo (approved):
This behavior seems reasonable.

PR #131, opened 2022-02-11
Added dotenv

CogniteJohan:


CogniteJohan:
Looks like a problem with state store

cognite/extractorutils/base.py
```diff
@@ -193,13 +195,23 @@ def __enter__(self) -> "Extractor":
...
             self
         """
+
+        # Environment Variables
+        self.env_file_path = find_dotenv()
+        if len(self.env_file_path) > 0:
+            load_dotenv(dotenv_path=self.env_file_path, override=True)
+            self.dotenv_message = f"Successfully ingested environment variables from\n{self.env_file_path}"
+        else:
+            self.dotenv_message = "Could not ingest environment variables from .env file"
```
mathialo, line 205: I suggest a slight rewording of this message, now it seems like something went wrong, while it's just that no env file exists (which I suspect might be the case for quite a few).

Maybe something like `No .env file located`?

cognite/extractorutils/base.py
```diff
@@ -193,13 +195,23 @@ def __enter__(self) -> "Extractor":
         Returns:
             self
         """
+
+        # Environment Variables
+        self.env_file_path = find_dotenv()
+        if len(self.env_file_path) > 0:
```
mathialo, line 201: To check if a string is not None and not empty, you can just use it as a boolean directly 🐍
``` suggestion
        if self.env_file_path:
```

cognite/extractorutils/base.py
mathialo: Also, if this message is only used here I suggest just using a local variable instead of assigning a new attribute (so no `self`) 🙂

mathialo:
I'm making a lot of mess for you now with merge conflicts - sorry! We're taking a few steps now to minimize those going forward, such as git-ignoring the lock file which has caused trouble for a while now.

Once I'm done with my current tasks, I'll take a proper look here and clean up my mess and see if we can get this PR merged soon 🙂

mathialo (approved):

PR #132, opened 2022-02-22
[CDF-14918] Base extensions class

einarmo:
Essentially just pulls the core code out of the MQTT and REST extensions and creates a common base extractor containing queues and the `handle_output` method. It isn't ideal to redefine types in all extension libraries, and it might be nice to have a place to put common behavior in the future.

### DIFF
--- a/cognite/extractorutils/_extensions.py
+++ b/cognite/extractorutils/_extensions.py
@@ -1,0 +1,191 @@
+#  Copyright 2022 Cognite AS
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""
+A module containing utilities for developing extensions for generic source systems.
+"""
+
+from datetime import datetime
+from types import TracebackType
+from typing import Callable, Iterable, List, Optional, Tuple, Type, Union
+
+from cognite.client import CogniteClient
+from cognite.client.data_classes import Event as _Event
+from cognite.client.data_classes import Row as _Row
+
+from cognite.extractorutils.configtools import CustomConfigClass
+from cognite.extractorutils.base import Extractor
+from cognite.extractorutils.statestore import AbstractStateStore
+from cognite.extractorutils.metrics import BaseMetrics
+from more_itertools import peekable
+
+from cognite.extractorutils.uploader import EventUploadQueue, RawUploadQueue, TimeSeriesUploadQueue
+
+try:
+    from typing import TypeAlias  # type: ignore
+except ImportError:
+    # Backport for python < 3.10
+    from typing_extensions import TypeAlias
+
+
+class RawRow:
+    def __init__(self, db_name: str, table_name: str, row: Union[_Row, Iterable[_Row]]):
+        self.db_name = db_name
+        self.table_name = table_name
+        if isinstance(row, Iterable):
+            self.rows = row
+        else:
+            self.rows = [row]
+
+
+TimeStamp = Union[int, datetime]
+
+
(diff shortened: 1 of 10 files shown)


cognite/extractorutils/_extensions.py
```diff
@@ -0,0 +1,187 @@
...
+        datapoints: Union[List[Tuple[TimeStamp, float]], List[Tuple[TimeStamp, str]]],
+    ):
+        self.id = id
+        self.external_id = external_id
+        self.datapoints = datapoints
+
+
+Event: TypeAlias = _Event
+
+CdfTypes = Union[Event, Iterable[Event], RawRow, Iterable[RawRow], InsertDatapoints, Iterable[InsertDatapoints]]
```
mathialo, line 69: These are a bit tricky. They will be needed by users of extensions, both to instantiate instances and for typing, so having them in a private module will yield all sorts of linting errors on every usage 🤔

tests/tests_unit/test_extensions_base.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+        ex.raw_queue = RawUploadQueue(client)
+
+        # Single
+        r = Row()
+        row = RawRow(db_name="some-db", table_name="some-table", row=r)
+        ex.handle_output(row)
+
+        ex.raw_queue.upload()
+        client.raw.rows.insert.ssert_called_with("some-db", "some-table", r)
```
mathialo, line 68: typo: `assert_called_with`

tests/tests_unit/test_extensions_base.py
```diff
@@ -0,0 +1,113 @@
...
+        # Iterable
+        r2 = Row()
+        rows = [
+            RawRow(db_name="some-db", table_name="some-table", row=r),
+            RawRow(db_name="some-db", table_name="some-table", row=r2),
+        ]
+        ex.handle_output(rows)
+
+        ex.raw_queue.upload()
+        client.raw.rows.insert.ssert_called_with("some-db", "some-table", [r, r2])
```
mathialo, line 79: typo: `assert_called_with`

tests/tests_unit/test_extensions_base.py
```diff
@@ -0,0 +1,113 @@
...
+            name="ext_extractor1", description="description", config_class=BaseConfig
+        )
+        ex.event_queue = EventUploadQueue(client)
+
+        # Single
+        evt = Event(external_id="some-event")
+        ex.handle_output(evt)
+
+        ex.event_queue.upload()
+        client.events.create.ssert_called_with([evt])
```
mathialo, line 44: typo: `assert_called_with`

tests/tests_unit/test_extensions_base.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Iterable
+        evts = [Event(external_id="some-event"), Event(external_id="some-other-event")]
+        ex.handle_output(evts)
+
+        ex.event_queue.upload()
+        client.events.create.ssert_called_with(evts)
```
mathialo, line 51: typo: `assert_called_with`

cognite/extractorutils/_extensions.py
```diff
@@ -0,0 +1,187 @@
...
+            config_file_path=config_file_path,
+            continuous_extractor=continuous_extractor,
+            heartbeat_waiting_time=heartbeat_waiting_time,
+            handle_interrupts=handle_interrupts,
+        )
+
+        self._event_queue_size = 10_000
+        self._raw_queue_size = 100_000
+        self._timeseries_queue_size = 1_000_000
+        self._upload_interval = 60
```
mathialo, line 130: These should either be configurable (by having a `ExtensionConfig` class that we use as base for the generic instead of `BaseConfig` with these in an `extractor` section or similar), or as module level constants IMO

mathialo (commented):
Nice, this is going to clean up the extension libraries by quite a bit :)

cognite/extractorutils/_extensions.py
mathialo: We could have every extension export them through a typing module just containing type aliases for these, but then it's yet another list to keep in sync with this

cognite/extractorutils/_extensions.py
einarmo: Good suggestion, a slightly extended config class somehow seems reasonable.

mathialo (approved):

mathialo (approved):

PR #134, opened 2022-03-03
fix: Loosen the requirements for dependencies

doctrino:
## Problem
The `cognite-extractor-utils` in incompatible with other Cognite sdks. For example, trying to install it with cognite-sdk-experimental leads to the following error message:

```
...And because cognite-extractor-utils (2.1.0) depends on typing-extensions (>=4.0.1,<5.0.0)
   and no versions of cognite-extractor-utils match >2.1.0,<3.0.0, cognite-sdk-experimental (>=0.60.33,<0.61.0) is incompatible with cognite-extractor-utils (>=2.1.0,<3.0.0).
  So, because common depends on both cognite-extractor-utils (^2.1.0) and cognite-sdk-experimental (^0.60.33), version solving failed
```

## Why

When you specify requirements using the carrot symbol, ^, you only go up to the next major release of the dependency, as seen above in the typing-extensions error message. https://python-poetry.org/docs/dependency-specification/#caret-requirements

## Solution

As this is a library, I prefer the requirements to be less strict and instead use >= such that the dependency resolution does not fail if any of the dependencies do a major version bump.


mathialo:
> And because cognite-extractor-utils (2.1.0) depends on typing-extensions (>=4.0.1,<5.0.0)

What's the beginning of that error message? What version of `typing-extensions` is the other library requiring? Looking [at the changelog](https://github.com/python/typing/blob/master/typing_extensions/CHANGELOG#L53), the breaking change from 3 to 4 was only removing certain things, so we might be able to go further back for that specific library.

We will not loosen up this widely, we need the guarantee for compatibility we get by locking at major releases, but we might be able to widen the interval for typing extensions back to `3.something`.

mathialo (approved):
Nice!

For later PRs, remember to update the changelog as well :)

doctrino:
> Nice!
> 
> For later PRs, remember to update the changelog as well :)

Will do, slipped my mind yesterday. Thanks for taking care of it.

PR #143, opened 2022-03-21
Allow time interval to be configured as strings

mathialo:
Let time intervals be configurable as `2m`, `30s` and so on. Implement
a simple parser that converts the number into seconds, and include
calculations to other common units.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,41 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+
+## Next
+
+### Changed
+
+ * Allow any interval to be configured as a string in addition to integers, so
+   e.g. upload intervals can be configured as
+
+   ```yaml
+   upload-interval: 2m
+   ```
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,32 @@ Changes are grouped as follows
...
+
+ * Allow any interval to be configured as a string in addition to integers, so e.g. upload intervals can be configured
+   as
+   ```yaml
+   upload-interval: 2m
+   ```
+   with `s`/`m`/`h`/`d` being valid units, and `s` being implied when units are missing (to preserve backwards 
+   compatibility with old config files). If your extractor reads from these config fields directly you need to update
+   to read the `seconds` (or the computed `minutes`, `hours` or `days` ) attribute instead of the fields directly, for
+   example:
```
mathialo, line 28: This is strictly speaking breaking. Although I expect most users will not read values directly like this, I think we might want to bump major versions on this one.

einarmo (approved):
Neat!

einarmo (approved):

PR #145, opened 2022-04-01
Improve error messages from invalid configs

mathialo:
Improve the exception messages generated in load_yaml. Instead of just
wrapping in an InvalidConfigError, make a custom error message. Handle
these error messages better in the base extractor class.

Some examples of new error messages:

```
$ poetry run python test.py test.yaml
Critical error: Could not read config file
Invalid config: Wrong type for field "source.test-number" - got "123"
of type str instead of int

$ poetry run python test.py test.yaml
Critical error: Could not read config file
Invalid config: Wrong type for field "source.test-number" - got "123"
of type str instead of int, float

$ poetry run python test.py test.yaml
Critical error: Could not read config file
Invalid config: Invalid YAML at line 22, column 21: mapping values are
not allowed here

$ poetry run python test.py test.yaml
Critical error: Could not read config file
Invalid config: Missing mandatory field "source"

$ poetry run python test.py test.yaml
Critical error: Could not read config file
Invalid config: Unknown config parameter "invalid-field"
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,6 @@ repos:
     -   id: isort
         additional_dependencies: [toml]
 -   repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
       - id: black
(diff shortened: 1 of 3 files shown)


einarmo (dismissed):
Neat!

PR #151, opened 2022-06-16
Catch errors in heartbeat

einarmo:
The extractor losing connection to CDF is no reason to kill the heartbeat, it may well be designed to handle it.


kjerand (approved):
Looks good

gaetan-h (approved):
LGTM

PR #157, opened 2022-08-18
Update cognite-sdk  to 4.0.1

emelvarol-cognite:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cognite-extractor-utils"
-version = "3.0.0-beta"
+version = "4.0.0-beta"
```
einarmo, line 3: I'd prefer `3.0.0-beta2`, or something along those lines. We're already on a beta of a major version, no need to skip 3.0.0 entirely.

emelvarol-cognite: Sure, I also wasn't sure about the versioning here. I guess it doesn't affect the users in a way that requires major version change. 🤔

emelvarol-cognite: Is it better now?

einarmo (approved):
LGTM, thanks

PR #158, opened 2022-08-18
Fix OAuthClientCredentials parameter usage

emelvarol-cognite:


emelvarol-cognite:
@einarmo sorry for the mistake here. They changed the naming for some oauth parameters in the new credential object. I missed the test was skipped and PR went through with this bug.

einarmo (dismissed):
Ew. This sucks. They should fix this in the SDK.

emelvarol-cognite:
> Ew. This sucks. They should fix this in the SDK.

I am keeping this on hold because of this issue: https://cognitedata.slack.com/archives/C010VCGKBCM/p1660833029009119
I'll bump the cog-sdk version after this is fixed and merge this pr (if ok 🤔 )

emelvarol-cognite:
> > Ew. This sucks. They should fix this in the SDK.
> 
> I am keeping this on hold because of this issue: https://cognitedata.slack.com/archives/C010VCGKBCM/p1660833029009119 I'll bump the cog-sdk version after this is fixed and merge this pr (if ok 🤔 )

I think this doesn't directly affect extractor-utils, maybe better to merge as the version on the master is broken now?

einarmo:
What I dislike about this particular patch is that I expect it will randomly break once they decide to align experimental and normal.

You can merge if you want, but we should keep an eye on it and fix it once they actually change this.

emelvarol-cognite:
> What I dislike about this particular patch is that I expect it will randomly break once they decide to align experimental and normal.
> 
> You can merge if you want, but we should keep an eye on it and fix it once they actually change this.

Hmm, what we could do now to be slightly safer, add the intended experimental sdk version to optional dependencies 🤔

einarmo:
> Hmm, what we could do now to be slightly safer is to add the intended experimental sdk version to optional dependencies 🤔

Sure, that's probably reasonable anyway. Right now we depend on it for remote configuration to work.

emelvarol-cognite:
> > Hmm, what we could do now to be slightly safer is to add the intended experimental sdk version to optional dependencies 🤔
> 
> Sure, that's probably reasonable anyway. Right now we depend on it for remote configuration to work.

I added the optional dependency, I am waiting on https://github.com/cognitedata/cognite-sdk-python/pull/1006/ to get merged. I'll ping you for approval after that.

emelvarol-cognite:
@einarmo since we use `cognite-sdk = "^4"` it will install the fixed version of cognite-sdk (4.1.1). No changes required. Can you review now?

einarmo (approved):
Looks good, thanks.

PR #161, opened 2022-09-16
Update experimental SDK

einarmo:
Now depends properly on the 4.x version of the SDK.


timurgen (approved):
lgtm, btw, pypi badge in sdk repo shows version 0.99.0, but pip seems to be ok to install 0.101.0

PR #163, opened 2022-09-28
[NOJIRA] added middleware setup with example JQ middleware

timurgen:
Adds middleware for UploaderExtractor which will call for each item in handle_output before putting it into a queue

### DIFF
--- a/cognite/extractorutils/uploader_extractor.py
+++ b/cognite/extractorutils/uploader_extractor.py
@@ -118,6 +118,7 @@ class UploaderExtractor(Extractor[UploaderExtractorConfigClass]):
 
         if isinstance(peek, Event):
             for event in peekable_output:
+                event = self._apply_middleware(event)
                 self.event_queue.add_to_upload_queue(event)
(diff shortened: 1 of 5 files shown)


cognite/extractorutils/uploader_extractor.py
```diff
@@ -129,6 +132,11 @@ def handle_output(self, output: CdfTypes) -> None:
         else:
             raise ValueError(f"Unexpected type: {type(peek)}")
 
+    def _apply_middleware(self, peek):
+        for mw in self.middleware:
+            peek = mw(peek)
+        return peek
```
einarmo, line 138: Minor thing, but I think `peek` here refers to looking at the contents of the iterator and deciding based on that, it's a little confusing to use it as a variable name in this method.

einarmo (dismissed):
This is neat. We could use this with a kind of "generic" input type to `handle_output` that applies the middleware before feeding it to upload queues.

einarmo (dismissed):

einarmo (approved):

PR #171, opened 2022-12-09
Added extraction pipelines decorator

CogniteJohan:
For old extractors it has proven very convenient to add an extraction pipelines decorator to add this functionality, so after the 2nd time I added this locally I thought it should be evaluated in the extractor utils library.

The logic is similar to the context manager for the extractor. One could consider refactoring so these will share some of the same base functions.

Unfortunately, I have not written any unittests as I do not know how to do this.

### DIFF
--- a/cognite/extractorutils/extraction_pipelines.py
+++ b/cognite/extractorutils/extraction_pipelines.py
@@ -1,0 +1,93 @@
+import logging
+import time
+from functools import wraps
+from threading import Thread
+
+from cognite.client import CogniteClient
+from cognite.client.data_classes import ExtractionPipelineRun
+
+
+def add_extraction_pipeline(
+    extraction_pipeline_ext_id: str,
+    cognite_client: CogniteClient,
+    logger: logging.Logger,
+    heartbeat_waiting_time: int = 5,
+    added_message: str = "",
+):
+    """
+    This is to be used as a decorator for extractor functions to add extraction pipeline information
+
+    Args:
+        extraction_pipeline_ext_id:
+        cognite_client:
+        logger:
+        heartbeat_waiting_time:
+        added_message:
+
+    Potential Improvements:
+    -- Refactor so this decorator share methods with the Extractor context manager in .base.py as they serve a similar
+    purpose
+    -- Add unit tests
+
+    """
+
+    def decorator_ext_pip(input_function):
+        @wraps(input_function)
+        def wrapper_ext_pip(*args, **kwargs):
+            ##############################
+            # Setup Extraction Pipelines #
+            ##############################
+            logger.info("Setting up Extraction Pipelines")
+
+            def _report_success() -> None:
+                message = f"Successful shutdown of function '{input_function.__name__}'. {added_message}"
+                cognite_client.extraction_pipelines.runs.create(
+                    ExtractionPipelineRun(external_id=extraction_pipeline_ext_id, status="success", message=message)
+                )
+
+            def _report_error(exception: Exception) -> None:
+                """
+                Called on an unsuccessful exit of the extractor
+                """
+                message = (
+                    f"Exception for function '{input_function.__name__}'. {added_message}:\n" f"{str(exception)[:1000]}"
+                )
+                cognite_client.extraction_pipelines.runs.create(
+                    ExtractionPipelineRun(external_id=extraction_pipeline_ext_id, status="failure", message=message)
+                )
+
+            def heartbeat_loop() -> None:
+                cognite_client.extraction_pipelines.runs.create(
+                    ExtractionPipelineRun(external_id=extraction_pipeline_ext_id, status="seen")
+                )
+
+                time.sleep(heartbeat_waiting_time)
+
+            ##############################
+            # Run the extractor function #
+            ##############################
+            logger.info(f"Starting to run function: {input_function.__name__}")
+
+            try:
+                heartbeat_thread = Thread(target=heartbeat_loop, name="HeartbeatLoop", daemon=True)
+                heartbeat_thread.start()
+                output = input_function(*args, **kwargs)
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


CogniteJohan:
@mathialo  my unittests passes the checks, but some integration tests seemingly unrelated to my changes are failing. Could you please have a look?

cognite/extractorutils/extraction_pipelines.py
```diff
@@ -0,0 +1,108 @@
...
+from threading import Event, Thread
+
+from cognite.client import CogniteClient
+from cognite.client.data_classes import ExtractionPipelineRun
+
+
+def add_extraction_pipeline(
+    extraction_pipeline_ext_id: str,
+    cognite_client: CogniteClient,
+    logger: logging.Logger = None,
```
mathialo, line 12: Passing in loggers like this is not something we do in the rest of the library, I think we should create a new one in here and use that. 

So add this at the top of the module and use it instead:

``` python
_logger = logging.getLogger(__name__)
```

cognite/extractorutils/extraction_pipelines.py
```diff
@@ -0,0 +1,108 @@
...
+            logger=<INSERT LOGGER>,
+        )
+        def extract_data(*args, **kwargs):
+            <INSERT FUNCTION BODY>
+
+    Potential Improvements:
+    -- Refactor so this decorator share methods with the Extractor context manager in .base.py as they serve a similar
+    purpose
+    -- Change 'cognite_client.extraction_pipeline_runs' -> 'cognite_client.extraction_pipelines.runs'
+    when SDK is updated
```
mathialo, line 41: Very good points for further improvements! But I suggest putting them in a normal comment with a `TODO` prefix instead of in the docstring as it is not particularly relevant to a user of the decorator :)

cognite/extractorutils/extraction_pipelines.py
```diff
@@ -0,0 +1,108 @@
...
+                    )
+
+                    cancellation_token.wait(heartbeat_waiting_time)
+
+            ##############################
+            # Run the extractor function #
+            ##############################
+            logger.info(f"Starting to run function: {input_function.__name__}")
+
+            print(cancellation_token.is_set())
```
mathialo, line 88: Avoid using `print` like this. If you need something logged you should use the logger instead, but I suspect this might be some debug printing that was just accidentally included?

cognite/extractorutils/extraction_pipelines.py
```diff
@@ -0,0 +1,108 @@
+import logging
```
mathialo, line 1: This file is missing copyright and license notice

mathialo (changes requested):
Very nice addition, and valuable insight on updating existing extractors! 

I have a few comments, but nothing major.

cognite/extractorutils/extraction_pipelines.py
CogniteJohan: Wouldn't it be convenient to pass your own logger so that it behaves similar to the rest of your code? However, I am all for conformity so I will be happy to change this, just want to check :)

cognite/extractorutils/extraction_pipelines.py
CogniteJohan: Absolutely! Will fix.

cognite/extractorutils/extraction_pipelines.py
CogniteJohan: Pure mistake on my part. Thanks for pointing it out. Will fix.

cognite/extractorutils/extraction_pipelines.py
CogniteJohan: Thanks, will add.

CogniteJohan:
Thanks for the comments, @mathialo . In addition to make the code base better, I also learn a lot from your comments as I am a non-programmer and have to pick up best practices etc along the way.

CogniteJohan:
@mathialo  Looks weird that the test fails for python 3.9 but not 3.8 nor 3.10. It does not look like it is caused by my changes as far as I can see.

CogniteJohan:
After merging in changes from master, it now fails for python 3.8 but works for 3.9 and 3.10...

cognite/extractorutils/extraction_pipelines.py
mathialo: IMO it's better to let frameworks define their own loggers, and configure the logging framework instead if you want control over what is actually logged from frameworks :)

cognite/extractorutils/extraction_pipelines.py
```diff
@@ -0,0 +1,116 @@
...
+from threading import Event, Thread
+
+from cognite.client import CogniteClient
+from cognite.client.data_classes import ExtractionPipelineRun
+
+
+def add_extraction_pipeline(
+    extraction_pipeline_ext_id: str,
+    cognite_client: CogniteClient,
+    heartbeat_waiting_time: int = 5,
```
mathialo, line 26: This is very short, in the base class it is set to `600` (ie, 10 minutes). Any reason it is set so low here? I think having the same default across the framework is nice.

Keep in mind that the backend will not check for missing heartbeats more than once every 15 minutes, so this will not make a huge difference in responsiveness

mathialo:
``` suggestion
    heartbeat_waiting_time: int = 600,
```

mathialo: Sorry for missing this the first time :)

cognite/extractorutils/extraction_pipelines.py
CogniteJohan: Agree. Thanks for pointing this out.

mathialo (approved):
Very nice!

PR #177, opened 2023-01-02
Catch errors when resolving config files

einarmo:


cognite/extractorutils/configtools.py
```diff
@@ -648,7 +648,11 @@ def is_remote(self) -> bool:
 
     @property
     def has_changed(self) -> bool:
-        self._resolve_config()
+        try:
+            self._resolve_config()
+        except Exception as e:
+            _logger.warning(f"Failed to reload configuration file: {str(e)}")
```
mathialo, line 654: If we used `_logger.exception` here instead we would not need to format the exception message, and we would get the entire stack trace in the logs, which could help a lot with debugging

mathialo (approved):
Consider changing log method

PR #178, opened 2023-01-06
Fix 'cancelation_token' typo throughout codebase

mathialo:
We had a typo where we asked for `cancelation_token` instead of
`cancellation_token`. Changing this is breaking, but since we are
breaking with the human readable time configs anyway, I want to make
this change too.


einarmo (approved):
Glorious

PR #181, opened 2023-01-13
Fix an issue with how we catch errors while reporting heartbeats

mathialo:
Reported by a user, they received

```
...

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/layers/google.python.runtime/python/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/layers/google.python.runtime/python/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/cognite/extractorutils/base.py", line 293, in heartbeat_loop
    except e:
NameError: free variable 'e' referenced before assignment in enclosing scope
```

which this should fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [4.0.1]
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


toondaey (approved):
lgtm

PR #199, opened 2023-03-22
Add support for certificate authentication

einarmo:


cognite/extractorutils/configtools.py
```diff
@@ -243,6 +249,33 @@ def __repr__(self) -> str:
...
+            cert_data, password=password.encode() if password else None
+        )
+        private_key_str = private_key.private_bytes(
+            encoding=serialization.Encoding.PEM,
+            format=serialization.PrivateFormat.TraditionalOpenSSL,
+            encryption_algorithm=serialization.NoEncryption(),
+        )
+        return (base64.b16encode(cert.fingerprint(hashes.SHA1())), private_key_str)
+    else:
+        raise InvalidConfigError("Unknown certificate format")
```
mathialo, line 276: Nit: maybe include what we think the format is, and what formats are allowed. This exception message is typically shown to the user, so it could help figuring out what's wrong

``` suggestion
        raise InvalidConfigError(f"Unknown certificate format '{path.suffix}'. Allowed formats are 'pem' and 'pfx'")
```

mathialo (commented):
Looks good, just one nitpick

mathialo (approved):

PR #201, opened 2023-04-11
Add `FileSizeConfig` class

mathialo:
Very similar to the existing `TimeIntervalConfig`. We implemented this in an extractor, and it might be useful elsewhere as well.

### DIFF
--- a/cognite/extractorutils/configtools.py
+++ b/cognite/extractorutils/configtools.py
@@ -249,6 +249,84 @@ class TimeIntervalConfig(yaml.YAMLObject):
         return self._expression
 
 
+class FileSizeConfig(yaml.YAMLObject):
+    def __init__(self, expression):
+        self._bytes, self._expression = FileSizeConfig._parse_expression(expression)
+
+    @classmethod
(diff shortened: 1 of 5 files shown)


cognite/extractorutils/configtools.py
```diff
@@ -249,6 +249,84 @@ def __repr__(self) -> str:
         return self._expression
 
 
+class FileSizeConfig(yaml.YAMLObject):
+    def __init__(self, expression):
+        self._bytes, self._expression = FileSizeConfig._parse_expression(expression)
```
toondaey, line 254: Maybe we should `expression` as `extension` instead.
```suggestion
        self._bytes, self._extension = FileSizeConfig._parse_expression(expression)
```

toondaey (changes requested):

cognite/extractorutils/configtools.py
mathialo: It's the whole expression, unaltered, and not just the unit :) It's stored to be returned back on `str(...)` for example.

toondaey (approved):

PR #202, opened 2023-04-18
Update raw queue sizes for uploader extractors

mathialo:
This change was made for db extractor etc before, but was missed here


rsjr (approved):
LGTM

PR #203, opened 2023-04-18
Repackage uploader and configtools

mathialo:
The modules have gotten big and messy, I think it's time to reorganize a bit. I _don't think_ I have broken anything, the tests still pass without any alterations. But maybe we should consider doing a major release anyway, both to be sure and to remove some of the deprecated stuff.


einarmo (approved):
LGTM!

PR #207, opened 2023-04-21
Breaking changes

mathialo:
A bundle of breaking changes. If we are breaking when updating the SDK, we might want to consider doing some other cleanups as well.

I have organized the PR into commits that hopefully makes sense as single pieces, meaning we can revert any of them before merging if we don't want to do it.

Mostly these fall into
 * Actually removing stuff marked as deprecated
 * Reducing the number of modules, several top-level modules were 50-line files with a single function or class in it, which is not a very pythonic way to organize a library.


rsjr (approved):
LGTM!

PR #217, opened 2023-06-11
feat: add global config parameters

nodegard:


cognite/extractorutils/__init__.py
```diff
@@ -16,5 +16,5 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "5.0.1"
+__version__ = "5.0.2"
```
einarmo, line 19: As this is a non-breaking feature update it should be a semimajor version, so we should be on `5.1.0`

cognite/extractorutils/configtools/elements.py
```diff
@@ -220,6 +236,7 @@ class CogniteConfig:
     data_set_external_id: Optional[str]
     extraction_pipeline: Optional[EitherIdConfig]
     timeout: TimeIntervalConfig = TimeIntervalConfig("30s")
+    global_config: GlobalConfig = field(default_factory=GlobalConfig)
```
einarmo, line 239: While it's called `global_config` in the SDK, I'm not so sure about exposing that to users who don't necessarily even know what kind of software the extractor is. Perhaps something like `connection`?

schema/cognite_config.schema.json
```diff
@@ -104,6 +104,59 @@
...
+        "global-config": {
+            "type": "object",
+            "description": "Configure authentication using OIDC tokens",
+            "properties": {
+                "disable-gzip": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Whether or not to disable gzipping of json bodies."
+                },
+                "disable-pypi-version-check": {
```
einarmo, line 117: This field isn't actually included in config, and shouldn't be.

einarmo (changes requested):
Looks good, I have a few things I'd like changed.

schema/cognite_config.schema.json
nodegard: Thanks for noticing, I'd added it before I noticed you were already setting that to a static `True` 👍

einarmo (approved):
LGTM. Thanks for doing this.

PR #234, opened 2023-08-02
Fix retry types and other issues with sequence uploader

mathialo:
Turns out mypy was right about complaining.

 * Fixed the type hint for the `retry` decorator. The list of exception types must be given as a tuple, not an arbitrary iterable.
 * Fixed retries for sequence upload queue.

### DIFF
--- a/cognite/extractorutils/uploader/_metrics.py
+++ b/cognite/extractorutils/uploader/_metrics.py
@@ -12,7 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from prometheus_client import Counter, Gauge, Histogram
+from prometheus_client import Counter, Gauge
 
 RAW_UPLOADER_ROWS_QUEUED = Counter(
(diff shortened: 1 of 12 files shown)


cognite/extractorutils/uploader/_metrics.py
```diff
@@ -24,22 +24,13 @@
     "cognite_raw_uploader_rows_duplicates", "Total number of duplicates found", labelnames=["destination"]
 )
 RAW_UPLOADER_QUEUE_SIZE = Gauge("cognite_raw_uploader_queue_size", "Internal queue size")
-RAW_UPLOADER_LATENCY = Histogram(
```
einarmo, line 27: Not clear to me why you removed these, it's something people will be asking for again.

mathialo: IIRC, it's something we added to prove something to one customer, and have been stuck with since. 

It is causing a bunch of headache, especially with typing. Which is why all the other builds are failing.

einarmo (approved):

PR #258, opened 2023-11-02
Fix environment variable resolution in remote check

einarmo:
We don't really need env. resolution in the remote check, but it will fail if you use the !env tag anywhere in the config file.


rsjr (approved):
the return of the aprob

mathialo (approved):

PR #260, opened 2023-11-09
add_iter_method_on_state_store

cel055:
added the method __iter__ on state store that will return the keys of the dict

### DIFF
--- a/cognite/extractorutils/statestore.py
+++ b/cognite/extractorutils/statestore.py
@@ -309,7 +309,7 @@ class AbstractStateStore(ABC):
     def __len__(self) -> int:
         return len(self._local_state)
 
-    def __iter__(self) -> Iterable:
+    def __iter__(self) -> Iterable[str]:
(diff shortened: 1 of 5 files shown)


cognite/extractorutils/statestore.py
```diff
@@ -309,6 +309,10 @@ def __contains__(self, external_id: str) -> bool:
     def __len__(self) -> int:
         return len(self._local_state)
 
+    def __iter__(self) -> Iterable:
```
mathialo, line 312: Include the type for the items as well
``` suggestion
    def __iter__(self) -> Iterable[str]:
```

mathialo (commented):
Would be great to see a test too :)

cognite/extractorutils/statestore.py
cel055: all set

mathialo (approved):

PR #262, opened 2023-11-13
Cognite-SDK v7

doctrino:


pyproject.toml
```diff
@@ -51,7 +51,7 @@ exclude = "tests/*"
 
 [tool.poetry.dependencies]
 python = "^3.8.0"
-cognite-sdk = ">=6.0, <7"
+cognite-sdk = "7.0.0a3"
```
haakonvt, line 54:
```suggestion
cognite-sdk = "^7"
```

mathialo (commented):
There is a version number in [`__init__.py`](https://github.com/cognitedata/python-extractor-utils/blob/master/cognite/extractorutils/__init__.py#L19C1-L19C22) as well which should be kept in sync.

But since this is already a breaking change we can just remove that, it's mostly a leftover from the pre-poetry days of the library. So you can also just remove it.

mathialo (approved):

PR #268, opened 2023-11-28
Azure Keyvault support

rsjr:
Adding Azure Keyvault support

### DIFF
--- a/tests/tests_unit/dummyconfig_keyvault.yaml
+++ b/tests/tests_unit/dummyconfig_keyvault.yaml
@@ -5,6 +5,7 @@ logger:
     level: INFO
 
 azure-keyvault:
+    authentication_method: client-secret
     client_id: ${KEYVAULT_CLIENT_ID}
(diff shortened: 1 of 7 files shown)


tests/tests_unit/dummyconfig_keyvault.yaml
```diff
@@ -0,0 +1,25 @@
...
+logger:
+  console:
+    level: INFO
+
+azure-keyvault:
+    authentication_method: client-secret
+    client_id: ${KEYVAULT_CLIENT_ID}
+    tenant_id: ${KEYVAULT_TENANT_ID}
+    client_secret: ${KEYVAULT_CLIENT_SECRET}
+    keyvault_name: extractor-credentials
```
mathialo, line 12: Some nitpick:

* Make these use hyphen-casing instead of snake_casing
* Rename `tenant_id` to `tenant` and `client_secret` to `secret` so that we use the same keys as in `idp-authentication`

mathialo (approved):

PR #271, opened 2023-12-11
feat(metrics): add option to specify dataset.

asahaf:
Add option to specify data set under which metrics timeseries will be created under when pushing metrics to cognite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [6.1.0]
+
+### Added
+
+ * Added ability to specify dataset under which metrics timeseries are created
+
 ## [6.0.1]
 
 ### Fixed
--- a/cognite/extractorutils/__init__.py
+++ b/cognite/extractorutils/__init__.py
@@ -16,5 +16,5 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "6.0.1"
+__version__ = "6.1.0"
 from .base import Extractor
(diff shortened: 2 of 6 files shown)


nrehmanawan (approved):

mathialo (changes requested):
Hi! And thanks for your contribution.

I think the linting will fail when it's run, but until then I think there are some style mistakes here. Run `pre-commit run --all` locally to fix them.

Also please update the changelog.

asahaf:
I've fixed the style and updated change log. However, pre-commit reports errors that are not related to the updated files/classes

mathialo:
The failing CI run is from the tests, and it's related to the updated files. This PR introduces a circular dependency which makes the tests (or anything else) incapable of importing anything from either module:

 * `cognite.extractorutils.metrics` depends on `cognite.extractorutils.configtools.elements` (imports `EitherIdConfig`)
 * `cognite.extractorutils.configtools.elements` depends on `cognite.extractorutils.metrics`  (it imports the pusher classes, of which `CognitePusher` uses `EitherIdConfig`)

This would cause an infinite recursion if the import system didn't catch it.

The `metrics` module can not depend on the `configtools.elements` module, neither should it. You should rewrite the change to the `CognitePusher` to take in an `EitherId` object instead of an `EitherIdConfig`, and in the  `start_pushers` method of `MetricsConfig`, you should initialize the `EitherId` from the `EitherIdConifg`. 

Hope that helps!

asahaf:
@mathialo replaced `EitherIdConfig` with `EitherId`. pre-commit passing now

mathialo:
Great!

The remaining failures seem to be related to some credentials needed for integration tests, which are not available to PRs from external repos. I'll merge this into an intermediate branch, and add a new PR from that branch.

asahaf:
Thanks @mathialo

mathialo:
When this merges, the feature should be released https://github.com/cognitedata/python-extractor-utils/pull/272

PR #272, opened 2023-12-12
Internal PR for #271

mathialo:


toondaey (approved):
Added a quick entry for some verbiage.

PR #273, opened 2023-12-13
Statestore: Retry on connection errors

amorken:
The SDK doesn't return a requests ConnectionError when it fails to connect, it returns a CogniteConnectionError. Retry on CogniteException to retry both CogniteAPIErrors and CogniteConnectionErrors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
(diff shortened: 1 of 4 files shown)


mathialo (commented):
Thanks!

There's a version number in `__init__.py` which also needs updating, and please add a little line in the changelog about the release :)

amorken:
I'd squash the last two commits into the former, but meh I never bothered cloning this I did it all in the github editor...   x_x

mathialo (approved):

PR #276, opened 2023-12-19
feat: custom success message for ExtractionPipelineRun

redzarosliCognite:
Modify Extractor base class for `success_message` option that is passed as an input to ExtractionPipelineRun() when a run is successful


einarmo (dismissed):
LGTM, thanks

einarmo (approved):

BugGambit (approved):

PR #284, opened 2024-01-19
Allow aliases for key vault config

mathialo:
To keep consistency with the dotnet variant, let's also allow to configure `key-vault` as an alias for `azure-keyvault` and `key-vault-name` as an alias for `keyvault-name`.

### DIFF
--- a/cognite/extractorutils/configtools/loaders.py
+++ b/cognite/extractorutils/configtools/loaders.py
@@ -61,7 +61,7 @@ class KeyVaultLoader:
                 "Include an `azure-keyvault` section in your config to use the !keyvault tag."
             )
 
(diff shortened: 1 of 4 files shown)


cognite/extractorutils/configtools/loaders.py
```diff
@@ -61,7 +61,8 @@ def _init_client(self) -> None:
                 "Include an `azure-keyvault` section in your config to use the !keyvault tag."
             )
 
-        if "keyvault-name" not in self.config:
+        keyvault_name = self.config.get("keyvault-name", self.config.get("key-vault-name"))
```
einarmo, line 64: Only `keyvault-name` is supported in .NET for now, so let's drop the other

```suggestion
        keyvault_name = self.config.get("keyvault-name")
```

einarmo (approved):

PR #289, opened 2024-02-06
Allow more flexible config of retry decorator

mathialo:
Option to specify retry exceptions as a dictionary instead of a tuple. Values should be a callable determining whether a specific exception object should be retied or not.

__Example:__

``` python
@retry(
   exceptions = {ValueError: lambda x: "Invalid" not in str(x)}
)
def func() -> None:
   value = some_function()

   if value is None:
       raise ValueError("Could not retrieve value")

   if not_valid(value):
       raise ValueError(f"Invalid value: {value}")
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,10 @@ Changes are grouped as follows
        value = some_function()
 
        if value is None:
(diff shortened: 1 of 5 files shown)


einarmo (approved):
Neat!

PR #292, opened 2024-02-13
Re-use the thread pool in file upload queues

mathialo:
Instead of creating a new threadpool in each upload call, re-use a single one.

Waiting on futures (without a timeout) will still make the the call to `upload()` blocking, so there should be no change in external usage.


einarmo (approved):
LGTM

PR #293, opened 2024-02-14
Change file upload queue to eager uploads

mathialo:
There is no real need for the file extractor to batch uploads like the other queue, in fact it makes it slower and worse.

This changes the file upload queues in some significant ways:

 - Instead of waiting to upload until a set of conditions, it starts uploading immediately.
 - The `upload()` method now acts more like a `join`, waiting on all the uploads in the queue to complete before returning.
 - A call to `add_to_upload_queue` when the queue is full will hang until the queue is no longer full before returning, instead of triggering and upload and hanging until everything is uploaded.
 - We require a max size, and remove the max wait time

As long as you use the queue in as a context, ie using
``` python
with FileUploadQueue(...) as queue:
```
and don't use time uploads, you should not have to change anything in your code. The behaviour of the queue will change, it will most likely be much faster, but it will not require any changes from you as a user of the queue.

I propose considering this a breaking change, and bumping the major version on next release, even though most users should be able to just update without changing their code. It's still a big change in the behaviour of the queue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## Next
+
+### Changed
+
+ * The file upload queues have changed behaviour.
+     - Instead of waiting to upload until a set of conditions, it starts uploading immedeately.
+     - The `upload()` method now acts more like a `join`, wating on all the uploads in the queue to complete before returning.
+     - A call to `add_to_upload_queue` when the queue is full will hang until the queue is no longer full before returning, instead of triggering and upload and hanging until everything is uploaded.
+     - The queues now require to be set up with a max size. The max upload latencey is removed.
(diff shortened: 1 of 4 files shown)


einarmo:
I know this works, but:

https://docs.python.org/3/library/threading.html#condition-objects

mathialo:
We could rewrite to using conditions, but you would still end up in a polling loop, this time checking the cancellation token. I did not find a neat way of both waiting on a condition and an event.

Maybe that's better? I guess we can poll much slower since triggering a shutdown happens less often and can safely wait a second or two.

einarmo:
> We could rewrite to using conditions, but you would still end up in a polling loop, this time checking the cancellation token. I did not find a neat way of both waiting on a condition and an event.
> 
> Maybe that's better? I guess we can poll much slower since triggering a shutdown happens less often and can safely wait a second or two.

Ugh, yeah, one would think an _event_ could be combined with other conditions somehow, but no.

einarmo (approved):

PR #294, opened 2024-02-15
Override `stop()` in file upload queues

mathialo:
Stop has taken on a different meaning, don't set the cancellation token on `stop()` for file upload queues


rsjr (approved):

mathialo:
With the new cancellation token, this is no longer that relevant.

PR #298, opened 2024-02-15
Cancellation token

einarmo:
This PR also enables the `B` class of lints in ruff, and fixes any warnings caused by that.

The cancellation token is a better abstraction than Event, since it allows hierarchical tokens, meaning we can cancel an upload queue without cancelling the entire extractor.


cognite/extractorutils/threading.py
```diff
@@ -0,0 +1,94 @@
...
+from threading import Condition
+from typing import Any, Optional
+
+
+class CancellationToken:
+    """
+    Abstraction for a hierarchical cancellation token.
+
+    Using this you can create hierarchies of cancellation tokens, to cancel a part of the extractor
+    without cancelling the whole process. Use `create_child_token` to create a token that will be
```
mathialo, line 12: Kind of nit, but python doc strings are restructured text and not markdown, so code snippets are double backticked
```suggestion
    without cancelling the whole process. Use ``create_child_token`` to create a token that will be
```

mathialo (approved):
Great!

PR #310, opened 2024-03-07
Support for files larger than 5GiB

einarmo:
Uses an abstraction for creating a wrapper around a stream that can create several smaller, sequential streams.


rsjr (approved):
LGTM

PR #312, opened 2024-03-12
make ConfigReloader not refresh auth token each cycle

haakonvt:
Hey, this is my first contribution to this repo, so if this PR looks crazy, please advise ❤️ 

 The problem I am trying to solve is the following:

By default, `config_refresher` runs every 5 minutes when a no-op `ReloadConfigAction` is given, i.e.:

```py
if self.reload_config_interval and self.reload_config_action != ReloadConfigAction.DO_NOTHING:
    Thread(target=config_refresher, name="ConfigReloader", daemon=True).start()
```

This code ends up calling `get_cognite_client` which forces a token refresh every time `config_refresher` runs:
```
config_refresher
-> self.config_resolver.has_changed
  -> _resolve_config
    -> tmp_config.cognite.get_cognite_client()
```

For some users using Auth0, every refresh incurs a cost... so doing this every 5 minutes VS 24 hours (default lifetime) actually accumulates into some 💲  💲  💲

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## 7.0.5
+
+### Fixed
+
+  * The background thread `ConfigReloader` now caches the `CogniteClient` to avoid repeatedly fetching a new token.
+
 ## 7.0.4
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


haakonvt:
CC @muradsater / @cognite-fholm

einarmo (approved):
This looks like a good change to me.

You might want to bump the version.

PR #329, opened 2024-05-16
Create a hashing state store

mathialo:
Create a state store which hashes the content and stores that, instead of a low/high watermark like the old version.

This will be useful for things like generic RAW rows and data models, where we don't always have an incremental field to use as a high watermark.


cognite/extractorutils/statestore.py
```diff
@@ -188,6 +178,36 @@ def synchronize(self) -> None:
...
+        cancellation_token: Optional[CancellationToken] = None,
+    ):
+        super().__init__(
+            save_interval=save_interval,
+            trigger_log_level=trigger_log_level,
+            thread_name=thread_name,
+            cancellation_token=cancellation_token,
+        )
+
+        self._local_state: Dict[str, Dict[str, Any]] = {}
```
ozangoktan, line 208: Is it feasible to use a type union here? I see the dictionary key "digest" being used in the AbstractHashStateStore, might be good to have typing.

ozangoktan (commented):
Might be a good idea to to split statestore.py into multiple files, there is 2 abstract bases now

mathialo:
> split statestore.py into multiple files

Ideally, I'd want to rename the old ones and reorganize a bit, but that would be breaking. I could try to split this up and keep the same behavior with some `__init__` magic

cognite/extractorutils/statestore.py
mathialo: Not without a breaking change.

ozangoktan:
> > split statestore.py into multiple files
> 
> Ideally, I'd want to rename the old ones and reorganize a bit, but that would be breaking. I could try to split this up and keep the same behavior with some `__init__` magic, but that won't fix the two base class issue.

Moving the hashing state stores to a submodule or another module at the same level would be preferable.

ozangoktan (approved):
LGTM 💯

PR #332, opened 2024-06-11
Fix loading keyvault parameters in remote config

einarmo:


mathialo (commented):
Change looks good, but should we combine with #333 and make a single release?

mathialo (approved):

PR #347, opened 2024-08-19
Upload with httpx

einarmo:
Upload the actual data stream to files using httpx instead of requests, to work around performance issue on windows.


rsjr (approved):
LGTM

PR #349, opened 2024-08-21
feat: split file upload

toondaey:


cognite/extractorutils/uploader/files.py
```diff
@@ -238,17 +243,111 @@ def _remove_done_from_queue(self) -> None:
...
+                mime_type=file_meta.mime_type,
+                uploaded_time=_timestamp_to_datetime(file_meta.uploaded_time)
+                if file_meta.uploaded_time is not None
+                else None,
+            )
+        )
+        node = instance_result.nodes[0]
+        return node.as_id()
+
+    def _upload_0(self, file_meta: FileMetadata, space: Optional[str] = None) -> FileMetadata:
```
mathialo, line 273: Nit: `_upload_empty` instead of `_upload_0`

cognite/extractorutils/util.py
```diff
@@ -501,3 +502,7 @@ def handle_cognite_errors(exception: CogniteException) -> bool:
         return True
 
     return {CogniteException: handle_cognite_errors}
+
+
+def _timestamp_to_datetime(ts: int) -> datetime:
```
mathialo, line 507: I can imagine this being useful in extractors as well, maybe just make it public?

While we're at it, should we add the inverse as well?

tests/tests_integration/test_file_integration.py
```diff
@@ -38,79 +40,145 @@ def set_test_parameters() -> ParamTest:
...
     # Upload the Filemetadata of an empty file without trying to upload the "content"
     queue.add_to_upload_queue(
-        file_meta=FileMetadata(external_id=test_parameter.external_ids[3], name=test_parameter.external_ids[3]),
+        meta_or_apply=FileMetadata(external_id=test_parameter.external_ids[3], name=test_parameter.external_ids[3]),
+        file_name=current_dir.joinpath("empty_file.txt"),
+    )
+
+    queue.add_to_upload_queue(
+        meta_or_apply=CogniteExtractorFileApply(
+            external_id=test_parameter.external_ids[0], name=test_parameter.external_ids[0], space=test_parameter.space
```
einarmo, line 89: Don't reuse external IDs in the tests. I realize it works, but it's very confusing, and the cleanup won't work like this. I'm reasonably certain you need to clean up these files in some special way.

toondaey: There's no alternative to external ID here because it is required by the instance. Working on cleaning up the files.

toondaey: Updated now.

mathialo (approved):

PR #360, opened 2024-09-12
Initial commit for new runtime and base class

mathialo:
The idea here is to split the responsibilities of the current base class into two: A runtime and an extractor.

The runtime is responsible for parsing command line arguments, loading config files and so on, before spawning the extractor in a separate process. The runtime will automatically restart the extractor if it crashes, but can also be asked by the extractor to restart it -  for example after a config change.

The extractor class is then only responsible for running the extractor application itself, making it much cleaner.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,43 +2,43 @@ name: release
 
 on:
   push:
-    branches: [ master ]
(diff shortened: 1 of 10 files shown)


rsjr (approved):
LGTM

PR #363, opened 2024-09-17
Scheduler implementation

mathialo:
Proposal for a new scheduler. Inspired by tha apscheduler, but with two important changes: Respecting a cancellation token both in the main loop and in sleeps, and allowing tasks to be triggered manually while still respecting the no overlap rule.

### DIFF
--- a/cognite/extractorutils/unstable/scheduling/_schedules.py
+++ b/cognite/extractorutils/unstable/scheduling/_schedules.py
@@ -1,5 +1,4 @@
 from abc import ABC, abstractmethod
-from random import randint
 from time import time
 
 from croniter import croniter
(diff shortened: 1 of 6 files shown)


cognite/extractorutils/unstable/scheduling/_schedules.py
```diff
@@ -0,0 +1,34 @@
...
+    def next(self) -> int:
+        return int(self._cron.get_next(start_time=time()))
+
+
+class IntervalSchedule(Schedule):
+    def __init__(self, interval: int) -> None:
+        self._interval = interval
+        self._jitter = randint(0, interval)
+
+        self._next = int(time() + self._jitter)
```
einarmo, line 27: I think jitter needs a rethink as discussed.

einarmo (commented):
Should maybe have some logging, and perhaps a test.

cognite/extractorutils/unstable/scheduling/_schedules.py
mathialo: Agree, I'll take it out for now and we can consider it later.

einarmo (approved):
Rocket Emoji

PR #364, opened 2024-09-17
Regex config options

toondaey:


cognite/extractorutils/configtools/elements.py
```diff
@@ -696,3 +696,39 @@ def create_state_store(
             return LocalStateStore(file_path="states.json", cancellation_token=cancellation_token)
         else:
             return NoStateStore()
+
+
+class RegExpFlag(Enum):
+    IGNORECASE = "ignore-case"
+    IC = "i"
+    ASCII = "ascii-only"
+    A = "a"
```
mathialo, line 705: Why two of each?

toondaey: For backward compatibility in the file extractor.

mathialo: Ah, I see. That's quite annoying. If we want to preserve backwards compatibility we'd have to change the `options` back to `flags` also I guess.

I don't think our current dataclass/dacite stack supports aliases, which is what we should have used here.

toondaey: We could support both `flags` and `options` at the same time but indicate which one will be deprecated in the next release cycle, maybe...

mathialo (approved):

PR #369, opened 2024-09-25
Fix UploadUrl when using privatelink 

rsjr:
https://cognitedata.atlassian.net/browse/DOG-4224

### DIFF
--- a/cognite/extractorutils/uploader/files.py
+++ b/cognite/extractorutils/uploader/files.py
@@ -408,12 +408,20 @@ class IOFileUploadQueue(AbstractUploadQueue):
         self, url_str: str, stream: BinaryIO, size: int, mime_type: Optional[str] = None
     ) -> Request:
         url = URL(url_str)
+        base_url = URL(self.cdf_client.config.base_url)
+
+        # same logic as the SDK
+        if url.netloc:
+            upload_url = url
+        else:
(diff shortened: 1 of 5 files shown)


cognite/extractorutils/uploader/files.py
```diff
@@ -408,12 +409,22 @@ def _get_file_upload_request(
...
     ) -> Request:
         url = URL(url_str)
+        base_url = URL(self.cdf_client.config.base_url)
+
+        if url.host == base_url.host:
+            upload_url = url
+        else:
+            parsed_url = urlparse(url_str)
+            parsed_base_url = urlparse(self.cdf_client.config.base_url)
+            replaced_upload_url = parsed_url._replace(netloc=parsed_base_url.netloc).geturl()
```
einarmo, line 419: Using an internal method is not ideal, is there no public alternative?

einarmo (commented):
I would love to see a tiny unit test, if only one that calls `_get_file_upload_request` directly.

cognite/extractorutils/uploader/files.py
rsjr: Nope. Tried to find it, but even in the urllib.parse docs they use the internal one as example =P 
https://docs.python.org/3/library/urllib.parse.html

einarmo (approved):

PR #371, opened 2024-10-02
Add task reporting to POC extractor class

mathialo:
Also add a test that uses parts of the checkin and startup parts of the base class.

### DIFF
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -1,14 +1,20 @@
 import logging
+from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import Queue
 from threading import RLock, Thread
 from types import TracebackType
 from typing import Generic, Literal, Optional, Type, TypeVar, Union
 
-from typing_extensions import Self
+from humps import pascalize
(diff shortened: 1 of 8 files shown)


cognite/extractorutils/unstable/core/base.py
```diff
@@ -40,29 +46,38 @@ def __init__(
...
         self.logger = logging.getLogger(f"{self.EXTERNAL_ID}.main")
 
     def _set_runtime_message_queue(self, queue: Queue) -> None:
         self._runtime_messages = queue
 
-    def _run_checkin(self) -> None:
-        def checkin() -> None:
-            body = {"externalId": self.connection_config.extraction_pipeline}
+    def _checkin(self) -> None:
+        with self._checkin_lock:
```
einarmo, line 60: Okay, a bit nitpicky, but I think we have to be careful about this stuff here:

If odin is not responding, this will block the entire extractor while retrying. What you _could_ do instead, is grab the checkin lock, empty the update list into a temporary, then release the lock. That way, the extractor will continue accumulating task updates while we retry against CDF.

einarmo (approved):

PR #372, opened 2024-10-02
Add support for castable integer and port number types in config schemas

ddonukis:
Previously it was not possible to specify a numeric field through environment variable substitution. For example a use case in the DB Extractor is to specify the port number of database connection using an environment variable.
```yaml
databases:
- name: test_db
  host: localhost
  port: ${DB_PORT_NUMBER}
```
The reason is that the libary used to convert raw yaml into data classes (`dacite`) does only rudimentary type validation and doesn't support trying to convert strings into expected types (like pydantic does for example).

A solution is to introduce a special subclass of `int` to be used in the config schema definition. This type is also added to the cast types in the `dacite.from_dict()`call so that the raw yaml value is passed to the constructor of the expected field type. This allows us to do some validation and also support both strings or numbers.

Initially I added `int` as a castable type in the yaml parser but it seemed like it broke some tests. And quite a lot of type can be implicitly converted to int with potentially unexpected results (e.g. converting booleans, truncating floats etc). Therefore, being explicit seemed like a better idea to me.

In order to benefit from the new feature the extractors depending on this package need to use the latest version and change the field type in the config schema definitions for instance:
```python
@dataclass(kw_only=True)
class OracleProviderConfig(BaseProviderConfig):
    type: Literal["oracle"]
    name: str
    user: str
    password: str
    host: str
    port: PortNumber = 1521
    service_name: Optional[str] = None
    timeout: Optional[CastableInt] = None
    batch_size: CastableInt = 1000
```

### DIFF
--- a/tests/tests_unit/test_configtools.py
+++ b/tests/tests_unit/test_configtools.py
@@ -295,18 +295,6 @@ def test_read_relative_path():
     assert config.path_field.name == "file.txt"
 
 
-def test_castable_int():
-    config = """
-    boolean-field: true
-    another-boolean-field: false
-    yet-another-boolean-field: false
-    string-field: "true"
-    another-string-field: "test"
-    yet-another-string-field: "test"
-    path-field: "./folder/file.txt"
-    """
-
-
 def test_read_absolute_path():
     config = """
     boolean-field: true
--- a/tests/tests_unit/test_configtools.py
+++ b/tests/tests_unit/test_configtools.py
@@ -34,7 +34,13 @@ from cognite.extractorutils.configtools import (
     load_yaml,
 )
 from cognite.extractorutils.configtools._util import _to_snake_case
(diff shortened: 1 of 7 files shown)


tests/tests_unit/test_configtools.py
```diff
@@ -289,6 +295,18 @@ def test_read_relative_path():
     assert config.path_field.name == "file.txt"
 
 
+def test_castable_int():
```
mathialo, line 298: Looks like this test isn't doing anything, maybe you forgot to add something?

mathialo (commented):
Very nice! Just one small question about a test

tests/tests_unit/test_configtools.py
ddonukis: Good catch! It was a first draft of a test that I forgot to remove.

ddonukis: Deleted it

mathialo (approved):
Thanks!

PR #384, opened 2024-11-10
Update README.md

tmolbach:
specified the wrong module. probably due to bad copy/paste.


einarmo:
Nope, this isn't a mistake.  `cogex` is a tool for creating extractors, it is installed as `cognite-extractor-manager`

PR #394, opened 2024-11-21
Changes for reporting files that have not been uploaded

nithinb:


cognite/extractorutils/uploader/upload_failure_handler.py
```diff
@@ -0,0 +1,58 @@
...
+    def add(self, file_name: str, error_reason: str) -> None:
+        error_file_object = FileErrorMapping(file_name=file_name, error_reason=error_reason)
+
+        self.failure_logs[FileFailureManager.FILE_REASON_MAP_KEY].update(error_file_object)
+
+        if len(self) >= self.MAX_QUEUE_SIZE:
+            self.write_to_file()
+            self.clear()
+
+    def write_to_file(self) -> None:
```
einarmo, line 53: I think this method should just clear internally, instead of leaving it to the caller.

nithinb: So I have now updated the write_to_file method to automatically trigger the clear. This should address this particular concern I believe

nithinb:
@einarmo I have addressed the changes on this, whenever you get a chance can you review it?

cognite/extractorutils/uploader/upload_failure_handler.py
```diff
@@ -0,0 +1,59 @@
...
+        return iter([[self.file_name, self.error_reason]])
+
+
+class FileFailureManager:
+    MAX_QUEUE_SIZE = 500
+    START_TIME_KEY = "start_time"
+    FILE_REASON_MAP_KEY = "file_error_reason_map"
+
+    def __init__(self, start_time: str | None = None, path_to_file: str | None = None) -> None:
+        self.failure_logs: dict[str, Any] = {}
```
einarmo, line 22: I don't really understand this implementation. From what I can tell, `failure_logs` only contains two fields `start_time` or `file_error_reason_map`. Why are these not attributes on the class instead of being hidden inside a dictionary without any typing?

nithinb: @einarmo So firstly I had to make it `Any` mainly because it causes issues while calling methods like `.update` which I call at a later point.

Also do you mean I should maintain it separately like `self.start_time` and `self.logs`. And before writing to the file just dynamically create a dictionary combining the logs and timestamp.

einarmo: Yeah, that would make better code, I think.

nithinb: okay that is fair. Fixed that change up @einarmo. let me know how it looks. 
Thanks in advance

pyproject.toml
```diff
@@ -75,6 +75,8 @@ httpx = "^0.27.0"
 pydantic = "^2.8.2"
 pyhumps = "^3.8.0"
 croniter = "^5.0.0"
+jsonlines = "^4.0.0"
+datetime = "5.5"
```
rsjr, line 79: datetime is a built-in lib, afaik? Not sure why it's listed/added here

cognite/extractorutils/uploader/upload_failure_handler.py
```diff
@@ -0,0 +1,64 @@
...
+        self._initialize_failure_logs()
+
+    def add(self, file_name: str, error_reason: str) -> None:
+        error_file_object = FileErrorMapping(file_name=file_name, error_reason=error_reason)
+        error_file_dict = dict(error_file_object)
+
+        self.failure_logs.update(error_file_dict)
+
+        if len(self) >= self.MAX_QUEUE_SIZE:
+            self.write_to_file()
```
rsjr, line 50: When it reaches the MAX_QUEUE_SIZE, it just flushes the content or appends the records to an existing file? People might try to reuse the same file over and over again, I guess

rsjr (commented):
Overall seems good, left some comments.

I know it might be a bit tricky, but it would be nice to have some tests describing the behavior and how the file must looks like end the end.

pyproject.toml
nithinb: oh okay let me remove this

cognite/extractorutils/uploader/upload_failure_handler.py
nithinb: so this is always going to be append-only to which ever file that is specified in the config file.

rsjr (approved):
LGTM. Let's merge this after the xmas break

PR #396, opened 2024-11-22
Add error reporting to integrations POC

mathialo:
Add an `error` method to the `Extractor` class that starts tracking of an error. This can be used in one of two ways, either by manually starting and stopping the error state:

``` python
e = extractor.error(...)

# handle error

e.finish()
```

or by using it as a context

``` python
with extractor.error(...):
    # Handle error
```

You can create an instant error (with no duration) by using the `instant()` method:

``` python
extractor.error(...).instant()
```

Tracking of start/end times, generating and keeping track of external IDs, reporting in checkins, and so on is all handled automatically.

### DIFF
--- a/cognite/extractorutils/unstable/configuration/models.py
+++ b/cognite/extractorutils/unstable/configuration/models.py
@@ -236,4 +236,4 @@ LogHandlerConfig = Union[LogFileHandlerConfig, LogConsoleHandlerConfig]
 
 
 class ExtractorConfig(ConfigModel):
-    log_handlers: List[LogHandlerConfig] = Field(default_factory=lambda: [LogConsoleHandlerConfig(level=LogLevel.INFO)])
+    log_handlers: List[LogHandlerConfig] = Field(default_factory=lambda: [LogConsoleHandlerConfig(level=LogLevel.INFO)])  # type: ignore[arg-type]
--- a/cognite/extractorutils/unstable/core/_dto.py
+++ b/cognite/extractorutils/unstable/core/_dto.py
@@ -7,6 +7,8 @@ from typing import Any, Literal
 from humps import camelize
 from pydantic import BaseModel, ConfigDict
 
+from cognite.extractorutils.unstable.core.errors import ErrorLevel
+
 
 class CogniteModel(BaseModel):
     """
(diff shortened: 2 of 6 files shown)


einarmo (approved):
This is probably fine for a start, but I think we should consider if there are ways to prevent people from accidentally reporting errors outside of a running task, since those would cause the checkin to fail.

mathialo:
> we should consider if there are ways to prevent people from accidentally reporting errors outside of a running task

This should do that, right? That was at least my intention with this approach. Now the task field is automatically picked up from the contextvar, which is set when entering a task call and unset when leaving. The user never sets it manually.

It's in the `add_task` method, we wrap the callable in two things: starting and ending tasks and adding and removing the context of being in a task (which is what sets the task field on errors), simplified without locks and error handling:

``` python
# start task
self._task_updates.append(
    TaskUpdate(type="started", name=task.name, timestamp=now()),
)

# add to context stack
context_token = self._current_task.set(task.name)

# Run task
target()

# unset context var
self._current_task.reset(context_token)

# end task
self._task_updates.append(
    TaskUpdate(type="ended", name=task.name, timestamp=now()),
)
```

And then the `error()` method uses this contextvar to detect whether we're in a task or not:

``` python
task_name = self._current_task.get()

return Error(
    level=level,
    description=description,
    details=details,
    extractor=self,
    task_name=task_name,
)
```

I can add some comments to the `add_task` method to clarify this if it wasn't clear already.

einarmo:
Oh! Neat, I didn't realize from looking at the code. Really nice. How does that handle threading? What happens if you run multiple tasks at the same time?

mathialo:
My understanding is that it acts like a call stack, so it should be fine. This type of behavior is what contextvars [are built for](https://docs.python.org/3/library/contextvars.html):

> Context managers that have state should use Context Variables instead of [threading.local()](https://docs.python.org/3/library/threading.html#threading.local) to prevent their state from bleeding to other code unexpectedly, when used in concurrent code.

 I did some testing with some simple threading where multiple threads, and it works as expected.

I'll add some comments to the `add_task` method, since this is clearly not obvious. 

All this said, I did get some 'the task is not running' errors when I left some test code running, so there might be some racyness I don't really see at the moment (or a bug in Odin). I'll have to investigate that further, but I'll merge this and leave that as a new thing afterwards.

PR #397, opened 2024-11-25
changes to set log level for HTTPX and HTTPCore

nithinb:


cognite/extractorutils/uploader/_base.py
```diff
@@ -56,8 +60,16 @@ def __init__(
...
         self.trigger_log_level = _resolve_log_level(trigger_log_level)
+
         self.logger = logging.getLogger(__name__)
 
+        log_level = logging.getLevelName(self.logger.getEffectiveLevel())
+
+        httpx_log_level = resolve_log_level_for_httpx(log_level)
+        httpx_logging_configuration = generate_config_for_http_logging(httpx_log_level)
+
+        logging.config.dictConfig(httpx_logging_configuration)  # type: ignore
```
einarmo, line 71: I really don't love this approach. It would be much better to actually build the config programatically, no? Here you just dump a dict and type: ignore it, we really should avoid that kind of logic whenever possible.

cognite/extractorutils/_inner_util.py
```diff
@@ -25,6 +25,45 @@ def _resolve_log_level(level: str) -> int:
...
 
 
+def resolve_log_level_for_httpx(level: str) -> str:
+    return {
+        None: "WARNING",
+        "INFO": "WARNING",
+        "WARNING": "WARNING",
+        "ERROR": "ERROR",
+        "CRITICAL": "CRITICAL",
+    }.get(level, "WARNING")
```
einarmo, line 35: We talked about setting the level to debug, I don't see any of that here?

This appears to _increase_ the level of all logging.

nithinb: >We talked about setting the level to debug, I don't see any of that here?

Yea sorry that was my bad. I hadn't pushed that change earlier

cognite/extractorutils/uploader/_base.py
nithinb: okay fair, can do. 

The reason for type: ignore is that mypy thinks logging.config is not a valid attribute.

nithinb: I found out that we are configuring all of the logger stuff in elements.py so was able to configure it as part of that. I think that should do the job. Do let me know if this looks good.

einarmo (approved):
LGTM!

PR #398, opened 2024-11-27
Improve exception handling in tasks

mathialo:
Log exception with stack trace, and optionally trigger a restart when something crashes.

### DIFF
--- a/cognite/extractorutils/unstable/core/restart_policy.py
+++ b/cognite/extractorutils/unstable/core/restart_policy.py
@@ -1,0 +1,29 @@
+from typing import Callable
+
+from cognite.extractorutils.unstable.core.tasks import ContinuousTask, Task
+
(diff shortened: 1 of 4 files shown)


cognite/extractorutils/unstable/core/restart_policy.py
```diff
@@ -0,0 +1,29 @@
+from typing import Callable
+
+from cognite.extractorutils.unstable.core.tasks import ContinuousTask, Task
+
+RestartPolicy = Callable[[Task], bool]
```
einarmo, line 5: Should this accept the exception as well? Might make sense to only restart on certain exceptions, maybe?

mathialo: good idea

einarmo (approved):
Neat!

PR #404, opened 2024-11-29
Improve runtime startup

mathialo:
Improve error handling for loading configs, roll back to previous versions if a version is invalid. Report errors back to integrations if a CDF-hosted application config fails to load.

Improve error handling overall, with better error messages relayed to the end user.

Do a connection check at startup to make sure the connection config is valid, with good error messages.

### DIFF
--- a/cognite/extractorutils/unstable/configuration/models.py
+++ b/cognite/extractorutils/unstable/configuration/models.py
@@ -139,7 +139,7 @@ class ConnectionConfig(ConfigModel):
     project: str
     base_url: str
 
-    extraction_pipeline: str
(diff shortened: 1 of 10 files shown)


einarmo (approved):

cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -139,7 +139,7 @@ class ConnectionConfig(ConfigModel):
     project: str
     base_url: str
 
-    extraction_pipeline: str
+    integration: str
```
eighty20results, line 142: 👌

PR #417, opened 2025-01-09
Loosen dependency restriction on httpx

mathialo:
Using `^0.x.y` in poetry is quite problematic, since it locks it to _minor_ version and only allows patch updates. This makes it very hard to depend on this package since it both makes it hard to update packages _and_ makes the likelihood of unresolvable requirements much higher.

The same is really true for upper bounds in general, but to a lesser extent. There's a quite good blog post on it from one of the big guys in scientific python:

https://iscinumpy.dev/post/bound-version-constraints/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ Changes are grouped as follows
 - `Security` in case of vulnerabilities.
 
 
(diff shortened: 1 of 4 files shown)


rsjr (approved):
👍

PR #423, opened 2025-01-20
Unstable: Update tasks and errors

mathialo:
See the changelog for a complete list of stuff that has changed

### DIFF
--- a/cognite/extractorutils/unstable/core/tasks.py
+++ b/cognite/extractorutils/unstable/core/tasks.py
@@ -1,4 +1,6 @@
+import logging
 from collections.abc import Callable
+from typing import TYPE_CHECKING
 
 from cognite.extractorutils.unstable.configuration.models import (
     CronConfig,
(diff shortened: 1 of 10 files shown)


cognite/extractorutils/unstable/core/tasks.py
```diff
@@ -1,30 +1,233 @@
...
+
+        self._logger = logging.getLogger(f"{self._extractor.EXTERNAL_ID}.{self._task.name.replace(' ', '')}")
+
+    def debug(self, message: str) -> None:
+        self._logger.debug(message)
+
+    def info(self, message: str) -> None:
+        self._logger.info(message)
+
+    def begin_warning(
```
einarmo, line 35: Idea: Could this be implemented as an interface-like that requires you to define `_error` and provides all the `begin_error`, etc. methods? There's some code duplication here that I think would be nice to avoid.

einarmo (commented):
I think this is mostly good, one thought on how to remove some duplication.

cognite/extractorutils/unstable/core/tasks.py
mathialo: that's a good idea. I was thinking about how to reuse the code here a bit, but was struggling to find some abstraction that made sense. Splitting out the `_error` method and making an abstract base class for the rest is nice, I like it.

einarmo (approved):
I like it!

PR #424, opened 2025-01-21
Add logging for LocalProtocolError for easy debuggability in the future

nithinb:
Adding logging to track errors such as `Too much data for declared Content` and `Too little data for declared Content-Length`. While there is no way for us to fix the issue but having better logging is always helpful to debug cases like these when customers face them.

### DIFF
--- a/cognite/extractorutils/uploader/files.py
+++ b/cognite/extractorutils/uploader/files.py
@@ -489,7 +489,7 @@ class IOFileUploadQueue(AbstractUploadQueue):
                     or "Too little data for declared Content-Length" in error_message
                 ):
                     self.logger.error(
-                        f"LocalProtocolError error while uploading file: {file_meta} error message: {error_message}"
+                        f"Content Length Mismatch while uploading file: {file_meta} error message: {error_message}"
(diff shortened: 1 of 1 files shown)


cognite/extractorutils/uploader/files.py
```diff
@@ -481,6 +482,19 @@ def wrapped_upload(
...
                 upload_file(read_file, file_meta)
 
+            except LocalProtocolError as e:
+                error_message = str(e)
+                if (
+                    "Too much data for declared Content" in error_message
+                    or "Too little data for declared Content-Length" in error_message
+                ):
+                    self.logger.error(
+                        f"LocalProtocolError error while uploading file: {file_meta} error message: {error_message}"
```
einarmo, line 492: While it's true that logging is for us, it is also for the user. I think this should be rephrased to be an informative message to the user. LocalProtocolError is an internal detail that isn't too interesting to the user IMO.

nithinb: okay that is a fair point, let me rephrase that a bit.

PR #425, opened 2025-01-29
Add a `from_environment` method to `ConnectionConfig`

mathialo:
Very useful in tests, and might also come in handy for docker builds


einarmo (approved):
Neat! We should add this to .NET as well.

PR #429, opened 2025-02-03
Changes to prevent the use of method find_dotenv

nithinb:
I have added information as part of the Jira ticket itself. 

Primarily we want to avoid the usage of `find_dotenv` since it recursively traverses to root in search of .env file. There is no way for us to limit this search. It is safer measure to stop looking if we don't find anything in the current directory. 

This will involve a major version bump primarily because I think this is a breaking change and needs to be communicated accordingly to customers as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## 7.6.0
+## 7.5.12
 
 ### Security
 
-* Removed reliance on `dotenv.find_dotenv` to prevent unintentional loading of environment files. Environment variable management is now more explicit to enhance security and reduce potential misconfigurations.
+* Disabled recursive search for .env file. The change will prevent loading environment variables hosted in directories other than the current working directory. Environment variable management is now more explicit to enhance security, and reduce potential misconfigurations.
 
 ## 7.5.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## 7.6.0
+
+### Security
+
+* Removed reliance on `dotenv.find_dotenv` to prevent unintentional loading of environment files. Environment variable management is now more explicit to enhance security and reduce potential misconfigurations.
+
 ## 7.5.11
 
 ### Added
--- a/cognite/extractorutils/__init__.py
+++ b/cognite/extractorutils/__init__.py
@@ -16,7 +16,7 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "7.6.0"
+__version__ = "7.5.12"
 from .base import Extractor
 
 __all__ = ["Extractor"]
--- a/cognite/extractorutils/__init__.py
+++ b/cognite/extractorutils/__init__.py
@@ -16,7 +16,7 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "7.5.11"
+__version__ = "7.6.0"
 from .base import Extractor
 
 __all__ = ["Extractor"]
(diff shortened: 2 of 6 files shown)


mathialo (commented):
This is a good change, but it is going to result in a breaking change in every extractor using the utils, so we need to be a bit careful when releasing this.

nithinb:
> This is a good change, but it is going to result in a breaking change in every extractor using the utils, so we need to be a bit careful when releasing this.

@mathialo  Yep I fully agree with this. I will include a version bump with valid release notes for the same. If the changes to address this look good then I will make the version changes as well.

nithinb:
@eighty20results I was hoping you could take a look at the security changelog. 
If it looks good I will reuse it with extractor PRs, to make the release

CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## 7.6.0
+
+### Security
+
+* Removed reliance on `dotenv.find_dotenv` to prevent unintentional loading of environment files. Environment variable management is now more explicit to enhance security and reduce potential misconfigurations.
```
eighty20results, line 19:
```suggestion
* Disabled recursive search for .env file to prevent use of a potential threat vector by loading environment variables hosted in directories above the current working directory. Environment variable management is now more explicit to enhance security and reduce potential misconfigurations.
```

cognite/extractorutils/__init__.py
```diff
@@ -16,7 +16,7 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "7.5.11"
+__version__ = "7.6.0"
```
eighty20results, line 19: Did we not agree to keep the version # change to a minor version (and call this a security update)?

CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## 7.6.0
```
eighty20results, line 15:
```suggestion
## 7.5.12
```
IIRC, there was a decision to keep the major version and "only" bump the minor version as this was going to be classified as a "minor update for security reasons" - Or did I misunderstand what we discussed during the stand-up @mathialo ?

eighty20results (commented):
See comments

cognite/extractorutils/__init__.py
nithinb: so I assumed since we are calling this a security update (rather than a bug-fix / patch), we should bump `7.5.X` to `7.6.0` since its more than just a bug fix/patch due to the security implications. 

Although I am happy to call it `7.5.12` since it might be an easier ask towards customers to make that upgrade.

nithinb:
@eighty20results the changes you requested have been made. 
Once this goes through, I will create the respective PRs for all the extractors and cut a release for them as well

mathialo (approved):

nithinb:
@mathialo  no issues at all. I have updated the version. If this looks good, we can get the maturity-review/risk-review also involved.

mathialo (approved):

larshagencognite:
Is it feasible to add testing for this?

mathialo:
@larshagencognite 
> Is it feasible to add testing for this?

In general, anything happening in the startup phase of extractors like this is __really__ hard to unit test, for many reasons such as

 * Parts of the startup is reading command line args and reacting to those. To test that would require a lot of monkeypatching.
 * Libraries like our SDK, but especially prometheus, keeps state under the hood hidden from users, making it very hard to call them several times independently, such as in different tests. For example, if you define a `Gauge` with the same name twice - as would happen if you called the extractor startup twice - the second call would fail.

With the new rewrite of the utils for next-gen extractors, we are working to decompose some parts of the extractor runtime to make single parts easier to test, but there will still be areas that are just not really possible to unit test without so much monkeypatching and mocking that the test itself becomes meaningless.

For changes like this we are much more dependent on manual testing and thorough reviews.

larshagencognite (approved):

PR #432, opened 2025-02-18
Change the completion message to something that looks more like success

ragnar-cognite:
- It shows up in CDF GUI

### DIFF
--- a/cognite/extractorutils/base.py
+++ b/cognite/extractorutils/base.py
@@ -218,7 +218,7 @@ class Extractor(Generic[CustomConfigClass]):
                 ExtractionPipelineRun(
                     extpipe_external_id=self.extraction_pipeline.external_id,
                     status="success",
-                    message="Successful shutdown",
+                    message="Successfully completed extraction",
                 )
             )
 


ragnar-cognite:
I tried to do a single line change just to see if I got the process right before I ventured into doing actual meaningful changes :-). 

What I really want is a dynamic message based on the outcome of the extractor run. So I guess something similar to the other PR you point to, except calling a function that will return a string so I can log stats as they look when we complete. I would pursue that as a separate PR.

But back to my single liner here: Are there really anyone out there that would prefer to get a success message saying something shut down? If I opened the GUI and saw that as my only message, I would think something went wrong and it died on me. I would also think part of the reason for using a framework like the extractor utils is to get a free ride for general improvements. Think of all the people out there we could make happy because CDF started telling them something went well, opposed to shutting down? :-).

The only thing I can think of breaking from such a chance is if somebody chain processing steps after each other and scan CDF logs looking for a shut down message. But if you rely on such brittle logic, you have a lot of bigger problems in the world. 

But I do realize I am new, and may miss something here. So please educate me and tell me where I went wrong!

mathialo:
>Are there really anyone out there that would prefer to get a success message saying something shut down?

This isn't as much about what message someone would _prefer_ to see in the front end, but what other systems are _depending_ on. Extraction pipelines is a service that has been used in many different ways by many different users, and several users have built other systems - such as their own monitoring and alerting applications that reads data from extraction pipelines. 

Changing what this value is for every extractor, which is what this PR would do, risk breaking any deployment where someone is depending on the content of messages. We're generally very weary of breaking any behavior in extractors outside of major releases.

ragnar-cognite:
OK, got it. Parking this. Might raise built in support for overriding completion messages again in the future.

PR #433, opened 2025-03-10
Handle case where input is a directory better

mathialo:
If a user configures a directory and not a simple file as the state store location, the error message produced is not very easy to understand. On windows in particular, it results in a "permission denied" error.

This adds a more descriptive error message


toondaey (approved):
LGTM

rbtcollins:
:unicorn:

PR #435, opened 2025-03-19
Attempt to make scheduler tests less flaky

mathialo:
Give the first run of the trigger test some time to finish, since the
scheduler doesn't allow multiple runs of the same task at the same time


einarmo (approved):

cemreyavuz (approved):
risk review ok

PR #437, opened 2025-05-05
Fix loading of prometheus server config

einarmo:
Loading the _PromServerConfig configuration object failed as it was not annotated as a dataclass, so it had no default conversion from a dict, which caused the YAML deserialization to fail.

Turns out this has _never_ been used.


nithinb (approved):

psalaberria002:
broken build, please fix

finnag:
Also please add a more thourough description. What was wrong, how did you fix it, why did that work?

psalaberria002 (approved):
:unicorn:

PR #438, opened 2025-05-07
Define _PromServerConfig as a dataclass

nithinb:


nithinb:
The tests on this will continue to fail but the change is unrelated to that. So I think it is still worth a review

einarmo:
#437  :)

nithinb:
> #437 :)

didn't see this. Approved that!!

PR #439, opened 2025-05-08
Add CDM Timeseries Upload Queue

devendra-lohar:
Add CDM Timeseries Upload Queue support for Timeseries Destination in DB Extractor.


cognite/extractorutils/uploader/time_series.py
```diff
@@ -340,6 +342,134 @@ def __len__(self) -> int:
...
+        datapoints = datapoints or []
+        old_len = len(datapoints)
+        datapoints = list(filter(self._is_datapoint_valid, datapoints))
+
+        new_len = len(datapoints)
+
+        if old_len > new_len:
+            diff = old_len - new_len
+            self.logger.warning(f"Discarding {diff} datapoints due to bad timestamp or value")
+            TIMESERIES_UPLOADER_POINTS_DISCARDED.inc(diff)
```
toondaey, line 410: Since this is the same as that of the base class, you could extract this into it's own method.

devendra-lohar: sure I can extract this as a method in the base itself.

devendra-lohar: commited the change.

cognite/extractorutils/uploader/time_series.py
```diff
@@ -340,6 +347,125 @@ def __len__(self) -> int:
...
+            max_delay=RETRY_MAX_DELAY,
+            backoff=RETRY_BACKOFF_FACTOR,
+        )
+        def _upload_batch(upload_this: list[dict]) -> list[dict]:
+            if len(upload_this) == 0:
+                return upload_this
+
+            self.cdf_client.time_series.data.insert_multiple(upload_this)
+
+            return upload_this
```
toondaey, line 441:
```suggestion
            if len(upload_this) > 0:
                self.cdf_client.time_series.data.insert_multiple(upload_this)

            return upload_this
```

cognite/extractorutils/util.py
```diff
@@ -79,35 +80,41 @@ def ensure_assets(cdf_client: CogniteClient, assets: Iterable[Asset]) -> None:
 
 class EitherId:
     """
-    Class representing an ID in CDF, which can either be an external or internal ID. An EitherId can only hold one ID
-    type, not both.
+    Class representing an ID in CDF, which can either be an external ID, internal ID or instance ID.
+    An EitherId can only hold one ID type, not more than one.
```
toondaey, line 84:
```suggestion
    An EitherId can only hold one ID type.
```

cognite/extractorutils/uploader/time_series.py
```diff
@@ -340,6 +347,123 @@ def __len__(self) -> int:
...
+            data_set_id,
+            cancellation_token,
+        )
+
+    def _apply_cognite_timeseries(self, timeseries_apply: CogniteExtractorTimeSeriesApply) -> NodeId:
+        instance_result = self.cdf_client.data_modeling.instances.apply(timeseries_apply)
+        node = instance_result.nodes[0]
+        return node.as_id()
+
+    def add_apply_to_upload_queue(
```
mathialo, line 395: This does not have the same signature as the parent class. That means that a `CDMTimeSeriesUploadQueue` type checks as a `TimeSeriesUploadQueue`, but doesn't behave like one.

A better approach here would be to make an abstract `BaseTimeSeriesUploadQueue` with all the common things, and have `TimeSeriesUploadQueue` and `CDMTimeSeriesUploadQueue` inherit from that.

mathialo: Quite surprised `mypy` didn't catch this :thinking:

mathialo: I see, the name is slightly different. 

That solves the polymorphism issue, but not the underlying issue that a CDM queue is not a subtype of a legacy queue. Now, the CDM queue has a `add_to_upload_queue` method which doesn't work. It would still be much better to factor out a separate base class and have the two queues implement their own `add_to_upload_queue` and `upload` methods

devendra-lohar: Makes sense I will create a BaseTimeSeriesUploadQueue and have the two inherit from this. I named it different based on the function behaviour but I see the point that it makes it deviate from the actual type.

cognite/extractorutils/uploader/time_series.py
```diff
@@ -83,7 +84,13 @@ def default_time_series_factory(external_id: str, datapoints: DataPointList) ->
     return TimeSeries(external_id=external_id, is_string=is_string)
 
 
-class BaseTimeSeriesUploadQueue(AbstractUploadQueue):
+def _apply_cognite_timeseries(cdf_client: CogniteClient, timeseries_apply: CogniteExtractorTimeSeriesApply) -> NodeId:
+    instance_result = cdf_client.data_modeling.instances.apply(timeseries_apply)
```
mathialo, line 88: The legacy equivalent does not create the time series in CDF itself, it only returns python objects that is then created in CDF by the queue.

I think we should try to keep the same behavior here

mathialo (commented):
The generic approach is pretty nice!

Noe more nitpick, but this is shaping up quite well

cognite/extractorutils/uploader/time_series.py
```diff
@@ -307,37 +394,186 @@ def _upload_batch(upload_this: list[dict], retries: int = 5) -> list[dict]:
...
+
+        @retry(
+            exceptions=cognite_exceptions(),
+            cancellation_token=self.cancellation_token,
+            tries=RETRIES,
+            delay=RETRY_DELAY,
+            max_delay=RETRY_MAX_DELAY,
+            backoff=RETRY_BACKOFF_FACTOR,
+        )
+        def _upload_batch(upload_this: list[dict], retries: int = 5) -> list[dict]:
```
mathialo, line 492: Please properly type your `dict`s

cognite/extractorutils/uploader/time_series.py
```diff
@@ -79,9 +84,15 @@ def default_time_series_factory(external_id: str, datapoints: DataPointList) ->
     return TimeSeries(external_id=external_id, is_string=is_string)
 
 
-class TimeSeriesUploadQueue(AbstractUploadQueue):
+def default_cdm_time_series_factory(
+    create_these_ids: set[NodeId], timeseries_apply_dict: dict[NodeId, CogniteExtractorTimeSeriesApply]
+) -> list[CogniteExtractorTimeSeriesApply]:
+    return [timeseries_apply_dict[nodeId] for nodeId in create_these_ids]
```
mathialo, line 90: This still works quite differently from the legacy TS queue.

The legacy queue takes an id and a list of datapoints in the `add_to_upload_queue` method, and calls the callback whenever it needs to go from id to full object (ie, when creating)

The new queue takes the full node object every call to add, in addition to the datapoints. And this callback does nothing.

The `add_to_upload_queue` method should take in a `NodeId` and a set of datapoints, and the callback should be called whenever the `CogniteExtractorTimeSeriesApply` actually needs to be generated.

mathialo (approved):
Looks good to me!

rsjr (commented):
This fix you added here https://github.com/cognitedata/python-extractor-utils/pull/439/commits/8e63e3661f69c2e9b21adb21517d7570bdb29da0

This doesn't look to be related to the changes in this PR? Is it fixing some other side effect from somewhere else?

rsjr (commented):
risk review ok

PR #440, opened 2025-05-21
Shorten the messages written to extractor pipeline if its more than 1000 characters

nithinb:
The method `self._report_run` is being defined here and will be removed from extractors where this is defined to make it easier to maintain.

### DIFF
--- a/cognite/extractorutils/base.py
+++ b/cognite/extractorutils/base.py
@@ -18,6 +18,7 @@ import sys
 from collections.abc import Callable
 from dataclasses import is_dataclass
 from enum import Enum
+from textwrap import shorten
 from threading import Thread
 from types import TracebackType
 from typing import Any, Generic, TypeVar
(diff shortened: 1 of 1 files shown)


einarmo:
Okay. The branch name on this PR is absurdly long. I'm not asking you to change it now, but please, keep branch names reasonable.

cognite/extractorutils/__init__.py
```diff
@@ -16,7 +16,7 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "7.5.14"
```
einarmo, line 19: This is a feature, should be a minor version bump, not a patch version.

cognite/extractorutils/base.py
```diff
@@ -210,37 +212,53 @@ def recursive_find_state_store(d: dict[str, Any]) -> StateStoreConfig | None:
...
-            message = f"{type(exception).__name__}: {str(exception)}"[:1000]
 
-            self.logger.info(f"Reporting new failed run: {message}")
-            self.cognite_client.extraction_pipelines.runs.create(
-                ExtractionPipelineRun(
-                    extpipe_external_id=self.extraction_pipeline.external_id, status="failure", message=message
-                )
-            )
+        if message is None and exception is None:
+            self.logger.exception("Failed to report error to Extraction Pipelines. No message or exception provided.")
```
einarmo, line 255: `logger.exception` is used to log python exceptions when in except blocks, you have no exception here to log.

cognite/extractorutils/base.py
```diff
@@ -210,37 +212,53 @@ def recursive_find_state_store(d: dict[str, Any]) -> StateStoreConfig | None:
...
-        if self.extraction_pipeline:
-            message = f"{type(exception).__name__}: {str(exception)}"[:1000]
 
-            self.logger.info(f"Reporting new failed run: {message}")
-            self.cognite_client.extraction_pipelines.runs.create(
-                ExtractionPipelineRun(
-                    extpipe_external_id=self.extraction_pipeline.external_id, status="failure", message=message
-                )
-            )
+        if message is None and exception is None:
```
einarmo, line 254: Actually, why do you have this check at all? `exception` is not nullable, as given by types. We do not generally add checks for types that are forbidden by python typing.

cognite/extractorutils/base.py
```diff
@@ -210,37 +212,53 @@ def recursive_find_state_store(d: dict[str, Any]) -> StateStoreConfig | None:
...
+        """
+        message = message or self.success_message
+        self._report_run("success", message)
+
+    def _report_error(self, exception: BaseException, message: str | None = None) -> None:
         """
         Called on an unsuccessful exit of the extractor
 
         Args:
             exception: Exception object that caused the extractor to fail
```
einarmo, line 251: You are missing doc comments for `message`. Also, you should probably make it clear that if `message` is missing, then `exception` will be used to create the error.

Also, perhaps, make it clear that if you provide a message, the exception is ignored. This isn't the greatest API, IMO.

cognite/extractorutils/base.py
```diff
@@ -210,37 +212,53 @@ def recursive_find_state_store(d: dict[str, Any]) -> StateStoreConfig | None:
 
         Extractor._statestore_singleton = self.state_store
 
-    def _report_success(self) -> None:
-        """
-        Called on a successful exit of the extractor
-        """
+    def _report_run(self, status: str, message: str) -> None:
```
toondaey, line 215: Maybe make the status more restricted.
```suggestion
    ReportStatus = Literal["success", "failure"]
    
    def _report_run(self, status: ReportStatus, message: str) -> None:
```

nithinb: This is a good idea, thanks.

cognite/extractorutils/base.py
nithinb: Naah you are right. 
Having two sources for message makes it confusing. I would rather split it up, if someone wants to report an error and they pass an exception, we construct the message. 
And if someone wants to directly provide the message and report a failure we should just have a _report_failure method that essentially creates a failure run.

CHANGELOG.md
```diff
@@ -12,12 +12,21 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## Next
```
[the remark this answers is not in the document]
nithinb: @toondaey Since I am releasing a version I have moved the content of next into the release version. I presume we only keep the section till we have things to add?

toondaey (approved):

PR #442, opened 2025-06-03
In `unstable`: Rename the local file override

mathialo:
`-l` / `--local-override` is renamed to `-f` / `--force-local-config` to
align with the dotnet version

Breaking changes are accepted in the `unstable` package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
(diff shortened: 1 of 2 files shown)


finnag (changes requested):
I think I am missing some context here, please explain.
Where is this change _used_? I would expect at least something else to change somewhere for the modified argument to have an effect, or be necessary.

mathialo:
* The change is in the `unstable` package, not used in any production extractors. It is so far only used in experiments.
* There is nowhere in this library where usage would change since this is only a library.

PR #444, opened 2025-06-05
Update connection config schema

toondaey:
This is to update the connection schema to [this](https://cognitedata.atlassian.net/wiki/spaces/DET/pages/4727635972/Config+design#Connection-config).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## 7.7.0
+
+### Added
+
+* Added the `CDMTimeSeriesUploadQueue` class to upload time series datapoints to instances in the Cognite Data Model (CDM). This new uploader uses `NodeId` for identification and can automatically create `CogniteExtractorTimeSeriesApply` instances if they are missing.
+
+### Changed
+
+* Refactored the time series uploaders for better code reuse and maintainability. A new generic base class, `BaseTimeSeriesUploadQueue`, has been introduced to contain common logic for datapoint validation and sanitization. Both the existing `TimeSeriesUploadQueue` and the new `CDMTimeSeriesUploadQueue` now inherit from this base class.
+
+
 ## 7.6.0
 
 ### Added
--- a/cognite/extractorutils/__init__.py
+++ b/cognite/extractorutils/__init__.py
@@ -16,7 +16,7 @@
 Cognite extractor utils is a Python package that simplifies the development of new extractors.
 """
 
-__version__ = "7.6.0"
+__version__ = "7.7.0"
 from .base import Extractor
 
 __all__ = ["Extractor"]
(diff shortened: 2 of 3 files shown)


cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -139,23 +147,40 @@ def __repr__(self) -> str:
...
-    proxies: dict[str, str] = Field(default_factory=dict)
+class RetriesConfig(ConfigModel):
+    max_retries: int = Field(default=10, ge=-1)
+    max_backoff: TimeIntervalConfig = Field(default_factory=lambda: TimeIntervalConfig("30s"))
     timeout: TimeIntervalConfig = Field(default_factory=lambda: TimeIntervalConfig("30s"))
 
 
+class SslCertificatesConfig(ConfigModel):
+    verify: bool = True
+    allow_list: list[str] | None = None
```
einarmo, line 158: I'm not sure this should allow the same space separation/comma separation syntax. It makes less sense here. We allow it for scopes for the specific reason I mentioned above, that it's supported by the OAuth standard.

cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -139,23 +147,40 @@ def __repr__(self) -> str:
...
+    @classmethod
+    def cast_thumbprints(cls, thumbprints: str | list[str] | None) -> list[str] | None:
+        if thumbprints is None:
+            return None
+        if isinstance(thumbprints, str):
+            return [scope.strip() for scope in thumbprints.split(",")]
+        return thumbprints
+
+
+class _ConnectionParameters(ConfigModel):
```
einarmo, line 170: I realize this is the same name as before, but I'm not sure this type should actually be private? It makes it impossible to safely construct a `ConnectionConfig` for users of the utils. What was the reasoning for this @mathialo?

cognite/extractorutils/unstable/configuration/models.py
toondaey: It's just a way of giving more options to users.

einarmo: I just don't think it's a good idea.

einarmo (approved):
I'm fine with these changes now.

cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -191,18 +213,26 @@ def __repr__(self) -> str:
...
-    proxies: dict[str, str] = Field(default_factory=dict)
+class RetriesConfig(ConfigModel):
+    max_retries: int = Field(default=10, ge=-1)
+    max_backoff: TimeIntervalConfig = Field(default_factory=lambda: TimeIntervalConfig("30s"))
     timeout: TimeIntervalConfig = Field(default_factory=lambda: TimeIntervalConfig("30s"))
 
 
+class SslCertificatesConfig(ConfigModel):
+    verify: bool = True
+    allow_list: list[str] | None = None
```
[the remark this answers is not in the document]
einarmo: A ticket to make the `allow_list` entry do something. Currently it is unused.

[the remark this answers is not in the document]
toondaey: Got it..

PR #445, opened 2025-06-05
Update `ruff` linting

mathialo:
Update the version of `ruff` running in CICD (as it was quite old).

Adding a few rules we have started having on a lot of our other codebases, and do the necessary changes in the code to comply with these rules.


einarmo (approved):

larshagencognite (approved):
🦄

PR #446, opened 2025-06-10
Add docstrings to every public mudule, class and function

mathialo:
This PR enables the `D` rule for ruff, which requires docstrings on every public module, class and function. It also contains some validation of the docstrings themeselves (such as ensuring that every function parameter is covered, and making sure the docstring adheres to the style guide.

I also intend on adding the `DOC` rule which adds further validation of the doc strings (such as looking for exceptions being thrown that aren't documented), but that also requires us to run in `preview` mode which will trigger more linting fixes, so I'll do that in a later PR.

I've also completely turned off the `D` rule for tests for now. While I think some simple documentation would be good for the tests, I'm trying to keep this PR small (believe it or not...).

This is a massive PR in terms of line changes, but it's only adding or fixing documentation strings, no functional changes are made.

### DIFF
--- a/cognite/extractorutils/base.py
+++ b/cognite/extractorutils/base.py
@@ -136,7 +136,7 @@ class Extractor(Generic[CustomConfigClass]):
         Load a configuration file.
 
         Either from the specified path, or from a path specified by the user in a command line arg. Will quit further
-        execution of no path is given.
+        execution if no path is specified.
 
         Args:
             override_path: Optional override for file path, ie don't parse command line arguments
--- a/cognite/extractorutils/configtools/_util.py
+++ b/cognite/extractorutils/configtools/_util.py
@@ -29,7 +29,7 @@ def _to_snake_case(dictionary: dict[str, Any], case_style: str) -> dict[str, Any
     """
     Ensure that all keys in the dictionary follows the snake casing convention.
 
-    This function will recursively fix any list or dictionaries.
+    This function will recursively fix any list or dictionaries. The input dictionary is never modified in place.
 
     Args:
         dictionary: Dictionary to update.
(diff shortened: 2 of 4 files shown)


toondaey (commented):
Not a part of this but also saw some errors:
1. [.cognite/extractorutils/exceptions.py](https://github.com/cognitedata/python-extractor-utils/blob/dde62e71ddd9db343e785f22900b12d98ed757d6/cognite/extractorutils/exceptions.py#L25) - "Unknown" instead of "Unkown".
2. [.cognite/extractorutils/metrics.py](https://github.com/cognitedata/python-extractor-utils/blob/dde62e71ddd9db343e785f22900b12d98ed757d6/cognite/extractorutils/metrics.py#L202) - "overriden" instead of "overrided".
3. [./cognite/extractorutils/unstable/configuration/exceptions.py](https://github.com/cognitedata/python-extractor-utils/blob/dde62e71ddd9db343e785f22900b12d98ed757d6/cognite/extractorutils/unstable/configuration/exceptions.py#L12) - "Unknown" instead of "Unkown".

Not sure if we can change this but also a spelling error:
[./cognite/extractorutils/uploader/time_series.py](https://github.com/cognitedata/python-extractor-utils/blob/dde62e71ddd9db343e785f22900b12d98ed757d6/cognite/extractorutils/uploader/time_series.py#L608) - "Sequence" instead of "Sequnce"

toondaey (dismissed):
👌🏾

thorkildcognite (commented):
Minor nits to consider

thorkildcognite:
risk review ok -- some small nits pointed out that you might fix, but nothing big.

PR #447, opened 2025-06-11
Fix type issue in `KeyValutLoader`

mathialo:
`SecretClient` does not accept `None`s as the `credential`. This is a
type issue, that `mypy` only _sometimes_ catches.

There is no reason the `credentials` should be a class variable, it's
only ever used inside the `_init_client` method, so it's fine to move it
to a local variable.


nithinb (approved):

larshagencognite (approved):
🦄

PR #452, opened 2025-07-02
[unstable] Add log level override 

devendra-lohar:
This PR adds support to override the log level defined in the application config from the cli.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## Next
+
+### Added
+
+* Added a new command-line argument, --log-level (-l) to allow users to override configured log levels for a single run.
+
 ## 7.7.0
 
 ### Added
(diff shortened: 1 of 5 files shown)


toondaey (commented):
Nice work. But I believe there's another `_setup_logging` in runtime.py that you may need to override.

devendra-lohar:
> Nice work. But I believe there's another `_setup_logging` in runtime.py that you may need to override.

The `_setup_logging` method in the Runtime class configures a very basic logger for the runtime process itself and is hardcoded to the `INFO` level to ensure initial startup messages are always visible.

My initial thought was to keep them separate to guarantee that critical startup messages are always visible. Let me know if I should adjust if a unified approach is preferred.

toondaey:
> > Nice work. But I believe there's another `_setup_logging` in runtime.py that you may need to override.
> 
> 
> 
> The `_setup_logging` method in the Runtime class configures a very basic logger for the runtime process itself and is hardcoded to the `INFO` level to ensure initial startup messages are always visible.
> 
> 
> 
> My initial thought was to keep them separate to guarantee that critical startup messages are always visible. Let me know if I should adjust if a unified approach is preferred.

You're right. The ticket does also say just log level in the extractor config.

toondaey (dismissed):

toondaey:
I just realized that we need an entry in the changelog for this.

rsjr (commented):
risk review ok

PR #455, opened 2025-07-04
Add ANN rule for ruff

mathialo:
Adding the annotation rule, which requires type annotations everywhere, and disallows certain shortcuts, like using `Any` to get around typing.

I had to be quite liberal with the exemptions, some of the fixes would be quite large and potentially breaking. Such as making the state stores generic over the type of state instead of just using `Any` everywhere.


einarmo:
:unicorn:

PR #458, opened 2025-07-16
Set start time of extractors.

toondaey:
This simply does nothing right now but just set the start time of the extractor. This will be used for one, when we report the startup.

### DIFF
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -47,6 +47,7 @@ The subclass should also define several class attributes:
 import logging
 import time
 from concurrent.futures import ThreadPoolExecutor
+from datetime import UTC, datetime
 from functools import partial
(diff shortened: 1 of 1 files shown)


cognite/extractorutils/unstable/core/base.py
```diff
@@ -435,6 +437,7 @@ def run(self) -> None:
             with extractor:
                 extractor.run()
         """
+        self._start_time = datetime.now(tz=timezone.utc)
```
thorkildcognite, line 440: nice to set it to UTC! Have had a ton of problems correlating when the customer has some funky timezones in use when running on their systems. Very confusing.

thorkildcognite:
risk review ok -- it will be good to report this in the future, indeed. Today, it is quite hard to detect extractors that end up in crash-looping (we've had that with the petrel studio connector for example -- although that is a very different type of beast that isn't in Python it taught me the need for this).

PR #460, opened 2025-07-18
Add example extractor to run unstable

devendra-lohar:
This PR adds a simple extractor example that logs messages at different levels and helps to run the unstable extractor for easy development.


devendra-lohar:
> This is good but could include the changes below, then in `pyproject.toml`, you can add somthing like:
> 
> ```toml
> [project.scripts]
> simple-extractor = "cognite.examples.unstable.extractors.simple_extractor:main"
> ```
> 
> That way we can always just do `uv run simple-extractor`

added this change.

paco-cognite:
risk review ok

PR #461, opened 2025-07-18
Move and update dtos

toondaey:
This is a preliminary work to properly managing how check-ins are reported for extractors. It involves moving some model objects as well as creating some new DTOs which do nothing for now but will be used in later PR(s).
This also moves when and where we set the start time for extractor to the `start` method as should be.

### DIFF
--- a/cognite/extractorutils/unstable/configuration/models.py
+++ b/cognite/extractorutils/unstable/configuration/models.py
@@ -8,7 +8,7 @@ from collections.abc import Iterator
 from datetime import timedelta
 from enum import Enum
 from pathlib import Path
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Literal, TypeVar
 
 from humps import kebabize
 from pydantic import BaseModel, ConfigDict, Field, GetCoreSchemaHandler
@@ -433,3 +433,7 @@ class ExtractorConfig(ConfigModel):
     """
 
     log_handlers: list[LogHandlerConfig] = Field(default_factory=_log_handler_default)
+
+
+ConfigType = TypeVar("ConfigType", bound=ExtractorConfig)
+ConfigRevision = Literal["local"] | int
--- a/cognite/extractorutils/unstable/core/_dto.py
+++ b/cognite/extractorutils/unstable/core/_dto.py
@@ -107,4 +107,4 @@ class CheckinRequest(HasExternalId):
 
 
 class CheckinResponse(HasExternalId):
-    last_config_revision: int | None
+    last_config_revision: int | None = None
--- a/cognite/extractorutils/unstable/core/_dto.py
+++ b/cognite/extractorutils/unstable/core/_dto.py
@@ -105,5 +105,5 @@ class CheckinRequest(WithExternalId):
     errors: ErrorList | None = None
 
 
-class CheckinResponse(HasExternalId):
+class CheckinResponse(WithExternalId):
     last_config_revision: int | None = None
--- a/cognite/extractorutils/unstable/core/_dto.py
+++ b/cognite/extractorutils/unstable/core/_dto.py
@@ -37,10 +37,14 @@ class HasExternalId(CogniteModel):
     external_id: str
 
 
+MessageType = Annotated[str, StringConstraints(min_length=0, max_length=1000)]
+
+
 class TaskUpdate(CogniteModel):
     type: Literal["started"] | Literal["ended"]
     name: str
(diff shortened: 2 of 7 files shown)


cognite/extractorutils/unstable/core/_dto.py
```diff
@@ -30,17 +33,78 @@ def dict(self, *args: Any, **kwargs: Any) -> dict[str, Any]:  # noqa: ANN401
...
     start_time: int
     end_time: int | None
     task: str | None
+
+
+TaskUpdateList = Annotated[list[TaskUpdate], Len(min_length=1, max_length=1000)]
+ErrorList = Annotated[list[Error], Len(min_length=0, max_length=1000)]
+VersionType = Annotated[str, StringConstraints(min_length=1, max_length=32)]
+DescriptionType = Annotated[str, StringConstraints(min_length=0, max_length=500)]
+TaskList = Annotated[list["Task"], Len(min_length=1, max_length=1000)]
```
ozangoktan, line 64: Will we enforce the min length in the api? Otherwise, maybe it could be better to leave it at 0.

cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -433,3 +433,7 @@ class ExtractorConfig(ConfigModel):
     """
 
     log_handlers: list[LogHandlerConfig] = Field(default_factory=_log_handler_default)
+
+
+ConfigType = TypeVar("ConfigType", bound=ExtractorConfig)
```
ozangoktan, line 438: There is an enum with the same name, we should have a different name for this.

toondaey: There's no other enum _within the module_ with the same name.

cognite/extractorutils/unstable/core/_dto.py
toondaey: These are enforced in the API hence these.

cognite/extractorutils/unstable/configuration/models.py
toondaey: I get that but with this new approach, we're eventually going to do away with the enum `ConfigType` because that will be handled by `ConfigRevision` instead.

ozangoktan (dismissed):

cognite/extractorutils/unstable/core/_dto.py
```diff
@@ -30,17 +33,77 @@ def dict(self, *args: Any, **kwargs: Any) -> dict[str, Any]:  # noqa: ANN401
...
+JSONType = TypeAliasType(  # type: ignore
+    "JSONType",
+    bool
+    | int
+    | float
+    | str
+    | None
+    | list[Optional["JSONType"]]  # type: ignore
+    | dict[str, Optional["JSONType"]],  # type: ignore  # type: ignore
+)
```
erlendvollset, line 73: For these type-ignores it's good practice to use `# type: ignore[specific-error-mode]`, both to indicate what is being ignored and also to avoid unintentionally ignoring different error modes.

cognite/extractorutils/unstable/core/_dto.py
```diff
@@ -30,17 +33,77 @@ def dict(self, *args: Any, **kwargs: Any) -> dict[str, Any]:  # noqa: ANN401
...
+TaskList = Annotated[list["Task"], Len(min_length=1, max_length=1000)]
+JSONType = TypeAliasType(  # type: ignore
+    "JSONType",
+    bool
+    | int
+    | float
+    | str
+    | None
+    | list[Optional["JSONType"]]  # type: ignore
+    | dict[str, Optional["JSONType"]],  # type: ignore  # type: ignore
```
erlendvollset, line 72:
```suggestion
    | dict[str, Optional["JSONType"]],  # type: ignore
```

cognite/extractorutils/unstable/core/errors.py
```diff
@@ -126,3 +127,15 @@ def __exit__(
         """
         self.finish()
         return exc_val is None
+
+    def into_dto(self) -> DtoError:
```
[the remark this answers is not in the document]
erlendvollset: We should also have a unit test for this mapper function.

ozangoktan (approved):

erlendvollset:
🦄

toondaey:
> https://github.com/cognitedata/pgpgpg/blob/main/.importlinter

Thank you! Will definitely consider it.

PR #464, opened 2025-07-25
Checkin worker for extractors

toondaey:


cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -0,0 +1,359 @@
...
+        self, cancellation_token: CancellationToken, startup_request: StartupRequest, interval: float | None = None
+    ) -> None:
+        with self._flush_lock:
+            while not cancellation_token.is_cancelled:
+                should_retry = self._report_startup(startup_request)
+                if not should_retry:
+                    self._has_reported_startup = True
+                    break
+                else:
+                    if not self._retry_startup:
```
einarmo, line 119: Nit: collapse this `else: if:` to an `elif`?

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -0,0 +1,359 @@
...
+                        raise RuntimeError("Could not report startup")
+
+                    a = (interval or STARTUP_BACKOFF_SECONDS) / 2
+                    b = (interval or STARTUP_BACKOFF_SECONDS) * 3 / 2
+                    next_retry = a + (b - a) * rng.random()
```
einarmo, line 124: `b - a` is equal to `a`, so you can just use that.

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -0,0 +1,359 @@
...
+    """
+    A worker to manage how we report check-ins to the Integrations API.
+
+    This will help us:
+    1. Ensure that we don't report any check-ins before the start up is reported.
+    1. Manage how we batch errors and task updates.
+    2. Manage how we handle retries and backoff.
+    """
+
+    _start_lock = RLock()
```
einarmo, line 45: The name `start_lock` is a little misleading, since it appears to be a lock around the `errors` and `task_updates` collections more than anything else.

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -0,0 +1,359 @@
...
+                self._errors.clear()
+                task_updates = self._task_updates[:]
+                self._task_updates.clear()
+
+            new_errors.sort(key=lambda e: e.end_time or e.start_time)
+            task_updates.sort(key=lambda t: t.timestamp)
+
+        while not cancellation_token.is_cancelled:
+            if len(new_errors) <= MAX_ERRORS_PER_CHECKIN and len(task_updates) <= MAX_TASK_UPDATES_PER_CHECKIN:
+                self._logger.debug("Writing check-in with no batching needed.")
```
einarmo, line 201: I think this log is misleading, since it will be hit if it's the _last_ batch being sent.

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -0,0 +1,359 @@
...
+            else:
+                new_errors = list(self._errors.values())
+                self._errors.clear()
+                task_updates = self._task_updates[:]
+                self._task_updates.clear()
+
+            new_errors.sort(key=lambda e: e.end_time or e.start_time)
+            task_updates.sort(key=lambda t: t.timestamp)
+
+        while not cancellation_token.is_cancelled:
```
einarmo, line 199: This is based on the logic in .NET utils, right? Maybe note that somewhere.

einarmo (commented):
This looks pretty good. I think some explanatory comments in the complex parts may be good, and I have some concerns around the locks, but it's taking shape. You will probably need a size exemption.

einarmo (dismissed):
I think this is pretty good. Link the equivalent .NET code in the PR description, and you will need to request a line count excemption.

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -0,0 +1,365 @@
...
+
+    def report_checkin(self, cancellation_token: CancellationToken) -> None:
+        """
+        Report a check-in to the Integrations API.
+
+        Arguments:
+        cancellation_token: A token to cancel the check-in reporting.
+        """
+        with self._lock:
+            if not self._has_reported_startup:
```
finnag, line 185: what situation leads to this scenario?

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -0,0 +1,365 @@
...
+                )
+            elif self._active_revision < checkin_response.last_config_revision:
+                self._logger.info(
+                    f"Remote config changed from {self._active_revision} to {checkin_response.last_config_revision}."
+                )
+                self._on_revision_change(checkin_response.last_config_revision)
+
+    def flush(self, cancellation_token: CancellationToken) -> None:
+        """
+        Flush available check-ins.
```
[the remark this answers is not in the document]
toondaey: The flush method will, in essence, upload check-in items i.e. `errors` and/or `task_events` if any is available. The term flush is used here since we're batch uploading the available items.

cognite/extractorutils/unstable/core/checkin_worker.py
toondaey: The `flush` method when fully implemented, can be called right when the [extractor](https://github.com/cognitedata/python-extractor-utils/blob/2302bc4f8f105cefb166b3c2ac35b48764850fa2/cognite/extractorutils/unstable/core/base.py#L409) is about to stop but there's a possibility that we're yet to actually report a start up for whatever reason, we would still want to report errors that aren't related to tasks. In that case, the block of code in this condition will filter out any task related `errors` and `task_events` before reporting the checkin.

finnag (approved):
:unicorn:

PR #470, opened 2025-08-18
[add] Implement state store in unstable module

devendra-lohar:
This PR aims to bring the state store related changes from base extractor to the unstable module. The state store provides a standardized way for extractors to persist state between runs, enabling efficient incremental extraction patterns.

### DIFF
--- a/cognite/extractorutils/unstable/configuration/models.py
+++ b/cognite/extractorutils/unstable/configuration/models.py
@@ -503,7 +503,7 @@ class StateStoreConfig(ConfigModel):
 
         if self.local:
             if self.local.path.is_dir():
-                raise ValueError(f"{self.local.path} is a directory, and not a file")
+                raise IsADirectoryError(f"{self.local.path}")
 
             return LocalStateStore(
                 file_path=str(self.local.path),
@@ -513,8 +513,8 @@ class StateStoreConfig(ConfigModel):
 
         if default_to_local:
             return LocalStateStore(file_path="states.json", cancellation_token=cancellation_token)
-        else:
-            return NoStateStore()
+
+        return NoStateStore()
 
 
 class ExtractorConfig(ConfigModel):
--- a/cognite/extractorutils/unstable/configuration/models.py
+++ b/cognite/extractorutils/unstable/configuration/models.py
@@ -503,7 +503,7 @@ class StateStoreConfig(ConfigModel):
 
         if self.local:
             if self.local.path.is_dir():
-                raise IsADirectoryError(f"{self.local.path}")
+                raise IsADirectoryError(self.local.path)
 
             return LocalStateStore(
                 file_path=str(self.local.path),
--- a/cognite/extractorutils/unstable/configuration/models.py
+++ b/cognite/extractorutils/unstable/configuration/models.py
@@ -5,6 +5,7 @@ Module containing pre-built models for common extractor configuration.
 import os
 import re
 from collections.abc import Iterator
+from dataclasses import dataclass, field
 from datetime import timedelta
 from enum import Enum
 from pathlib import Path
@@ -24,6 +25,13 @@ from cognite.client.credentials import (
 )
 from cognite.extractorutils.configtools._util import _load_certificate_data
 from cognite.extractorutils.exceptions import InvalidConfigError
+from cognite.extractorutils.statestore import (
+    AbstractStateStore,
+    LocalStateStore,
+    NoStateStore,
+    RawStateStore,
+)
+from cognite.extractorutils.threading import CancellationToken
 
 __all__ = [
     "AuthenticationConfig",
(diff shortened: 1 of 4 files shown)


toondaey (commented):
The state store should also be passed as a task context because it is meant to be used by tasks to keep track of their states.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -245,6 +256,63 @@ def _setup_logging(self) -> None:
...
+
+        try:
+            self.state_store.initialize()
+        except ValueError:
+            self._logger.exception("Could not load state store, using an empty state store as default")
+
+        Extractor._statestore_singleton = self.state_store
+
+    @classmethod
+    def get_current_statestore(cls) -> AbstractStateStore:
```
toondaey, line 302: We should have a test for this as well.

toondaey (dismissed):
Looks good!

cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -427,11 +434,95 @@ def _log_handler_default() -> list[LogHandlerConfig]:
...
+                cdf_client=cdf_client,
+                database=self.raw.database,
+                table=self.raw.table,
+                save_interval=self.raw.upload_interval.seconds,
+                cancellation_token=cancellation_token,
+            )
+
+        if self.local:
+            if self.local.path.is_dir():
+                raise ValueError(f"{self.local.path} is a directory, and not a file")
```
haakonvt, line 506:
```suggestion
                raise IsADirectoryError(self.local.path)
```

cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -427,11 +434,95 @@ def _log_handler_default() -> list[LogHandlerConfig]:
...
+            return LocalStateStore(
+                file_path=str(self.local.path),
+                save_interval=self.local.save_interval.seconds,
+                cancellation_token=cancellation_token,
+            )
+
+        if default_to_local:
+            return LocalStateStore(file_path="states.json", cancellation_token=cancellation_token)
+        else:
+            return NoStateStore()
```
haakonvt, line 517: Code 🏌️ 
```suggestion
        return NoStateStore()
```

cognite/extractorutils/unstable/core/base.py
```diff
@@ -245,6 +256,63 @@ def _setup_logging(self) -> None:
...
+
+        def recursive_find_state_store(d: dict[str, Any]) -> StateStoreConfig | None:
+            for k in d:
+                if is_dataclass(d[k]):
+                    res = recursive_find_state_store(d[k].__dict__)
+                    if res:
+                        return res
+                if isinstance(d[k], StateStoreConfig):
+                    return d[k]
+            return None
```
haakonvt, line 278: This short-circuits after the first found? Is there any scenario when you would have multiple/ambiguous and this should raise?

cognite/extractorutils/unstable/core/base.py
```diff
@@ -245,6 +256,63 @@ def _setup_logging(self) -> None:
...
+        """
+
+        def recursive_find_state_store(d: dict[str, Any]) -> StateStoreConfig | None:
+            for k in d:
+                if is_dataclass(d[k]):
+                    res = recursive_find_state_store(d[k].__dict__)
+                    if res:
+                        return res
+                if isinstance(d[k], StateStoreConfig):
+                    return d[k]
```
haakonvt, line 277: Suggestion:
```suggestion
            for k, v in d.items():
                if is_dataclass(v):
                    res = recursive_find_state_store(v.__dict__)
                    if res:
                        return res
                if isinstance(v, StateStoreConfig):
                    return v
```

cognite/extractorutils/unstable/core/base.py
```diff
@@ -245,6 +256,63 @@ def _setup_logging(self) -> None:
...
+                cdf_client=self.cognite_client,
+                default_to_local=self.USE_DEFAULT_STATE_STORE,
+                cancellation_token=self.cancellation_token,
+            )
+        else:
+            self.state_store = (
+                LocalStateStore("states.json", cancellation_token=self.cancellation_token)
+                if self.USE_DEFAULT_STATE_STORE
+                else NoStateStore()
+            )
```
haakonvt, line 292: I find this easier to read, but just a suggestion
```suggestion
        elif self.USE_DEFAULT_STATE_STORE:
            self.state_store = LocalStateStore("states.json", cancellation_token=self.cancellation_token)
        else:
            self.state_store = NoStateStore()
```

cognite/extractorutils/unstable/core/base.py
```diff
@@ -245,6 +256,63 @@ def _setup_logging(self) -> None:
...
+
+        try:
+            self.state_store.initialize()
+        except ValueError:
+            self._logger.exception("Could not load state store, using an empty state store as default")
```
haakonvt, line 297: Don't you need an additional `self.state_store = NoStateStore()` ?

tests/test_unstable/conftest.py
```diff
@@ -202,3 +202,14 @@ def log_messages_task(ctx: TaskContext) -> None:
             ctx.warning("This is a warning message.")
 
         self.add_task(StartupTask(name="log_task", target=log_messages_task))
+
+
+@pytest.fixture(autouse=True)
+def reset_singleton() -> Generator[None, None, None]:
```
haakonvt, line 208: It's possible to use the slightly shorter notation:
```suggestion
def reset_singleton() -> Iterator[None]:
```

I would also put this `autouse=True` fixture at the top of the file since it runs before and after every single test case and thus should be very visible 😄 

Last note, I would typically reach for `monkeypatch.setattr` to temporarily override attributes for tests (but no need to use here, as you also want to nuke it as part of test teardown).

haakonvt (changes requested):
A few questions from risk review, mostly LGTM 😄

tests/test_unstable/conftest.py
devendra-lohar: added this

cognite/extractorutils/unstable/core/base.py
devendra-lohar: makes sense

cognite/extractorutils/unstable/core/base.py
devendra-lohar: thank you for pointhing this out. 
This implementation could lead to the pick up of some other StateStoreConfig incase there are multiple. I have chnaged it to directly pick the state_store in the Extractor Base Config.

haakonvt:
🦄

haakonvt:
🦄

PR #475, opened 2025-09-05
Add Metrics and pusher to unstable

Hmnt39:
This PR introduces the metrics and pusher manager for Cognite extractors, enabling flexible metrics collection, singleton enforcement, and reliable pushing to Prometheus and Cognite Data Fusion (CDF).

### DIFF
--- a/cognite/extractorutils/unstable/configuration/models.py
+++ b/cognite/extractorutils/unstable/configuration/models.py
@@ -411,6 +411,15 @@ class LogLevel(Enum):
     INFO = "INFO"
     DEBUG = "DEBUG"
 
+    @classmethod
+    def _missing_(cls, value: object) -> "LogLevel":
+        if not isinstance(value, str):
+            raise ValueError(f"{value} is not a valid log level")
+        for member in cls:
+            if member.value == value.upper():
+                return member
+        raise ValueError(f"{value} is not a valid log level")
+
 
 class LogFileHandlerConfig(ConfigModel):
     """
--- a/tests/test_unstable/test_configuration.py
+++ b/tests/test_unstable/test_configuration.py
@@ -7,6 +7,7 @@ from cognite.client.credentials import OAuthClientCredentials
 from cognite.extractorutils.unstable.configuration.loaders import ConfigFormat, load_io
 from cognite.extractorutils.unstable.configuration.models import (
     ConnectionConfig,
+    LogLevel,
     TimeIntervalConfig,
     _ClientCredentialsConfig,
 )
@@ -212,3 +213,14 @@ def test_from_env() -> None:
 
     # Check that the produces cogniteclient object is valid
     assert len(client.assets.list(limit=1)) == 1
+
+
+def test_setting_log_level_from_any_case() -> None:
+    log_level = LogLevel("DEBUG")
+    assert log_level == LogLevel.DEBUG
+
+    log_level = LogLevel("debug")
+    assert log_level == LogLevel.DEBUG
+
+    with pytest.raises(ValueError):
+        LogLevel("not-a-log-level")


einarmo:
Is the metrics module just copied? Do you need to change it? Any reason why you can't just reuse the one already in the stable part of the utils?

Hmnt39:
> Is the metrics module just copied? Do you need to change it? Any reason why you can't just reuse the one already in the stable part of the utils?

Aren't we moving everything to unstable ? Should I use the stable `configtools` and `metrics` in unstable?

cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -429,6 +435,176 @@ class LogConsoleHandlerConfig(ConfigModel):
...
+
+class _PushGatewayConfig(ConfigModel):
+    """
+    Configuration for pushing metrics to a Prometheus Push Gateway.
+    """
+
+    host: str
+    job_name: str
+    username: str | None
+    password: str | None
```
einarmo, line 467: Won't the lack of defaults here means that these config objects will need an explicit `username: null` in config?

Could you write a test that actually loads these types from YAML?

tests/test_unstable/test_base.py
```diff
@@ -228,3 +235,98 @@ def test_raw_state_store_integration(
...
+        "push_gateways": [
+            {
+                "host": "http://localhost:9091",
+                "job_name": "test-job",
+                "clear_after": None,
+            }
+        ],
+        "cognite": None,
+        "server": None,
+    }
```
einarmo, line 312: Are you dumping a struct as YAML then reading it back? This doesn't really verify that `None` values aren't still required. Instead, you should write out the yaml and just load it directly from a string in the test.

tests/test_unstable/test_base.py
```diff
@@ -298,30 +298,26 @@ def counting_push(self: "PrometheusPusher") -> None:
...
-        yaml.dump(yaml_config, f)
+def test_pushgatewayconfig_none_credentials_from_yaml() -> None:
+    config_str = """
+push-gateways:
+  - host: "http://localhost:9091"
+    job_name: "test-job"
+    clear_after: null
+    username: null
+cognite: null
+server: null
```
einarmo, line 309: The thing I wanted to verify was that this works without explicitly setting `x: null`, so leave those out. We generally don't encourage explicitly setting things to null in yaml configs.

Hmnt39: Here, password is not set explicitly to null, so this is working (I want to verify if explicitly setting to null is also working i.e. username).

Hmnt39: > The thing I wanted to verify was that this works without explicitly setting `x: null`, so leave those out. We generally don't encourage explicitly setting things to null in yaml configs.

Made the changes, I got your point why we do require to set the defaults and not passing null explicitly.

cognite/extractorutils/unstable/configuration/models.py
```diff
@@ -429,6 +435,176 @@ class LogConsoleHandlerConfig(ConfigModel):
...
+                push_interval=push_gateway.push_interval.seconds,
+                thread_name=f"MetricsPusher_{counter}",
+                cancellation_token=self.cancellation_token,
+            )
+            prometheus_pusher.start()
+            self.pushers.append(prometheus_pusher)
+            if push_gateway.clear_after is not None:
+                self.clear_on_stop[prometheus_pusher] = push_gateway.clear_after.seconds
+
+        if self.metrics_config.cognite:
```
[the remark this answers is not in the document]
Hmnt39: Added the tests

PR #479, opened 2025-09-12
Fix a potential unbound variable

toondaey:
This is a simple fix, however, not trivial as this could result in a potentially unbound variable.


einarmo (dismissed):
This is the kind of thing switching to strict pyright would have caught.

toondaey:
> This is the kind of thing switching to strict pyright would have caught.

True

toondaey:
> As discussed on Slack, it makes sense to add some tests here. Ping me when you are done so I can re-review it!

The test has been update now.

rsjr (approved):
risk review ok

PR #480, opened 2025-09-15
Add fallback method for encoding support

Hmnt39:
This PR improves file reading by adding a fallback mechanism for encoding.
Previously, attempting to read files that were not encoded in `UTF-8` would raise a `UnicodeDecodeError`. With this change, the code will now:

1. Attempt to read files using `UTF-8`.
2. If `UTF-8` decoding fails, automatically fall back to the system’s locale-preferred encoding (e.g., CP1252 on Windows).


einarmo (approved):
In PRs like this, you should include _why_ this is necessary in the PR description.

Hmnt39:
> Member

I'll keep that in mind.

PR #516, opened 2026-03-24
Add context to validation

toondaey:
This makes it easy for us to pass context data to pydantic validation. One of it's use cases is that of the test where we can then update this context thereby allowing nested models to validate based on the value provided by the parent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## 7.12.1
+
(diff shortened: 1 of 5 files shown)


Hmnt39 (dismissed):
Looks good to me

trygveu (dismissed):
LGTM, I have a couple of suggestions to make it clear that this is Pydantic context. (Context is a slightly overloaded term)

trygveu (approved):
Thanks
risk review ok

Hmnt39 (approved):

trygveu (approved):
Risk review ok

PR #518, opened 2026-04-06
Fix log message for Loading app config from CDF

Yaseen-A-Khan:
The log showed repeated "Loading application config from CDF" when the config in the integrations UI is invalid.
If same error were to repeat, it wasn't logged.

This PR:
- logs the invalid error message along with the "Loading application config from CDF" so that the user has more context why there is trouble in loading the config from CDF

### DIFF
--- a/cognite/extractorutils/unstable/core/logger.py
+++ b/cognite/extractorutils/unstable/core/logger.py
@@ -343,7 +343,6 @@ class RobustFileHandler(TimedRotatingFileHandler):
             errors=errors,
         )
 
-        # FileHandler leaves stream unset until first emit when delay=True; typeshed types it as optional.
-        if self.stream is not None:
-            self.stream.write("")
-            self.stream.flush()
+        # Ensure the log file path exists and is writable. Using touch validates when
+        # delay=True, because FileHandler does not open the stream until the first emit.
+        filename.touch()
--- a/cognite/extractorutils/unstable/core/logger.py
+++ b/cognite/extractorutils/unstable/core/logger.py
@@ -343,5 +343,7 @@ class RobustFileHandler(TimedRotatingFileHandler):
             errors=errors,
         )
 
-        self.stream.write("")
-        self.stream.flush()
+        # FileHandler leaves stream unset until first emit when delay=True; typeshed types it as optional.
+        if self.stream is not None:
+            self.stream.write("")
+            self.stream.flush()
--- a/cognite/extractorutils/unstable/core/runtime.py
+++ b/cognite/extractorutils/unstable/core/runtime.py
@@ -298,7 +298,9 @@ class Runtime(Generic[ExtractorType]):
                 raise InvalidConfigError(str(e)) from e
 
         else:
-            self.logger.info("Loading application config from CDF")
+            self.logger.info(
+                f"Loading application config from CDF for integration: {connection_config.integration.external_id}"
+            )
 
             application_config, current_config_revision = load_from_cdf(
                 self._cognite_client,
--- a/cognite/extractorutils/unstable/core/runtime.py
+++ b/cognite/extractorutils/unstable/core/runtime.py
@@ -333,9 +333,10 @@ class Runtime(Generic[ExtractorType]):
 
             except Exception as e:
                 error_message = str(e)
-                self.logger.error(f"{error_message}")
+                self.logger.error(error_message)
                 if error_message == prev_error:
-                    # Wait and retry if the error is the same as the previous one
+                    # Back off with a random delay in [1, RETRY_CONFIG_INTERVAL] seconds so retries are spread over time
+                    # to reduce synchronized retries / thundering herd when loading config fails.
                     self._cancellation_token.wait(randint(1, self.RETRY_CONFIG_INTERVAL))
                     continue
                 prev_error = error_message
--- a/cognite/extractorutils/unstable/core/runtime.py
+++ b/cognite/extractorutils/unstable/core/runtime.py
@@ -333,12 +333,11 @@ class Runtime(Generic[ExtractorType]):
 
             except Exception as e:
                 error_message = str(e)
+                self.logger.error(f"{error_message}")
                 if error_message == prev_error:
-                    # Log the error again to make sure it is not missed
-                    self.logger.error(error_message)
+                    # Wait and retry if the error is the same as the previous one
                     self._cancellation_token.wait(randint(1, self.RETRY_CONFIG_INTERVAL))
                     continue
-                self.logger.error(error_message)
                 prev_error = error_message
 
                 ts = now()
--- a/cognite/extractorutils/unstable/core/runtime.py
+++ b/cognite/extractorutils/unstable/core/runtime.py
@@ -334,7 +334,8 @@ class Runtime(Generic[ExtractorType]):
             except Exception as e:
                 error_message = str(e)
                 if error_message == prev_error:
-                    # Same error as before, no need to log it again
+                    # Log the error again to make sure it is not missed
+                    self.logger.error(f"{error_message}")
                     self._cancellation_token.wait(randint(1, self.RETRY_CONFIG_INTERVAL))
                     continue
                 self.logger.error(error_message)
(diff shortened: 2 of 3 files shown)


devendra-lohar (commented):
mypy checks are failing for this

cognite/extractorutils/unstable/core/runtime.py
```diff
@@ -334,7 +334,8 @@ def _safe_get_application_config(
             except Exception as e:
                 error_message = str(e)
                 if error_message == prev_error:
-                    # Same error as before, no need to log it again
+                    # Log the error again to make sure it is not missed
+                    self.logger.error(f"{error_message}")
```
devendra-lohar: I don't understand why would we want to log this again. I think we should improve on the message `Loading application config from CDF`

Yaseen-A-Khan: As on every retry, the function is called, the message "Loading application config from CDF" is logged to terminal, 
logging error along with the the loading message during the retries would provide more context about what is occurring during the load.

devendra-lohar (dismissed):
LGTM

cognite/extractorutils/unstable/core/runtime.py
```diff
@@ -334,7 +334,8 @@ def _safe_get_application_config(
             except Exception as e:
                 error_message = str(e)
                 if error_message == prev_error:
-                    # Same error as before, no need to log it again
+                    # Log the error again to make sure it is not missed
+                    self.logger.error(error_message)
                     self._cancellation_token.wait(randint(1, self.RETRY_CONFIG_INTERVAL))
```
nithinb, line 340: Did you get a chance to look at why we do a random wait instead of a fixed wait?

cognite/extractorutils/unstable/core/runtime.py
```diff
@@ -334,7 +334,8 @@ def _safe_get_application_config(
             except Exception as e:
                 error_message = str(e)
                 if error_message == prev_error:
-                    # Same error as before, no need to log it again
+                    # Log the error again to make sure it is not missed
+                    self.logger.error(error_message)
```
nithinb: It is a good thing that we are logging the error message again, but is there a chance of over-logging this?

nithinb (changes requested):
Also what would be the new log message when a config is not available in the Integrations UI? 

Currently, one of the problems is that, if there is no Application Config available on the Integrations UI, it would log the message Loading Application Config which is not fully accurate since there is no Application Config available.

cognite/extractorutils/unstable/core/runtime.py
```diff
@@ -334,7 +334,8 @@ def _safe_get_application_config(
             except Exception as e:
                 error_message = str(e)
                 if error_message == prev_error:
-                    # Same error as before, no need to log it again
+                    # Log the error again to make sure it is not missed
+                    self.logger.error(f"{error_message}")
                     self._cancellation_token.wait(randint(1, self.RETRY_CONFIG_INTERVAL))
                     continue
                 self.logger.error(error_message)
```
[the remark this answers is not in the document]
nithinb: I think its worth addressing this comment @Yaseen-A-Khan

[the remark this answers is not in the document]
Yaseen-A-Khan: Yes, this approach reduces the duplication while keeping the code flow simple.

cognite/extractorutils/unstable/core/runtime.py
Yaseen-A-Khan: The intention is to log the error message again everytime the `try_get_application_config()` is called to load the application config from CDF, so that the error message doesnt get buried in the "Loading application config from CDF" log.

Yaseen-A-Khan: and going by the flow of function calls, the "Loading application config from CDF" will be logged before the error message is logged.
So, if there were an error, it can be thought of as a pair - Loading message followed by error message.

cognite/extractorutils/unstable/core/runtime.py
Yaseen-A-Khan: I have raised a query with Mathias and am awaiting his response.

Yaseen-A-Khan: @mathialo could you please provide insight as to why .wait() using randint(1, 30) (RETRY_CONFIG_INTERVAL = 30) is being applied ?

cognite/extractorutils/unstable/core/runtime.py
nithinb: fair

cognite/extractorutils/unstable/core/runtime.py
nithinb: Since you started a thread in #dog-team-huginn, based on the outcome of that discussion we can decide if we should change that.

nithinb (dismissed):
🐴

cognite/extractorutils/unstable/core/runtime.py
Yaseen-A-Khan: Absolutely

trygveu: This gives you jitter on retries, the cognite recommended pattern is truncated exponential backoff with jitter. See: https://docs.cognite.com/dev/concepts/resource_throttling

cognite/extractorutils/unstable/core/runtime.py
```diff
@@ -333,11 +333,11 @@ def _safe_get_application_config(
 
             except Exception as e:
                 error_message = str(e)
+                self.logger.error(f"{error_message}")
```
trygveu: nit(non-blocking): You might want to tighten up the ruff rules for this repository. A tiny optimization that might not be worth doing is https://docs.astral.sh/ruff/rules/logging-f-string/ 

Just good to be aware of.

cognite/extractorutils/unstable/core/logger.py
```diff
@@ -343,5 +343,7 @@ def __init__(
             errors=errors,
         )
 
-        self.stream.write("")
-        self.stream.flush()
+        # FileHandler leaves stream unset until first emit when delay=True; typeshed types it as optional.
+        if self.stream is not None:
+            self.stream.write("")
+            self.stream.flush()
```
trygveu: The intention of this stream write is likely to verify that the filepath is valid and writeable. If the logger is initialized with delay=True, we're loosing that check with this change. I believe it would we more pythonic and direct to just verify that the file is created and writeable by:
```suggestion
            filename.touch()  # Verify that we have a valid path and permissions to write the logfile.
```

trygveu (commented):
I have a nit on the init that I'd like to hear your thoughts on. And when you're touching the code, you might want to add an explanatory comment on the jitter retry since that was a topic of conversation in this review. 

Really good work!

cognite/extractorutils/unstable/core/logger.py
Yaseen-A-Khan: I believe, this helps for the check right at the `init` without checking for stream.

cognite/extractorutils/unstable/core/runtime.py
Yaseen-A-Khan: Thank you that was helpful. In the interest of existing consistency, I believe eliminating the f-string and going with `self.logger.error(error_message)` would be right

cognite/extractorutils/unstable/core/logger.py
```diff
@@ -343,5 +343,6 @@ def __init__(
             errors=errors,
         )
 
-        self.stream.write("")
-        self.stream.flush()
+        # Ensure the log file path exists and is writable. Using touch validates when
+        # delay=True, because FileHandler does not open the stream until the first emit.
+        filename.touch()
```
nithinb, line 348: Nit: This doesn't block the PR itself, but something to keep in mind when we do this change

if we do a `filename.touch()` we are technically not respecting the case when `RobustFileHandler` is called with delay=True. This doesn't affect us currently since in our [code](https://github.com/cognitedata/python-extractor-utils/blob/c1b05f20e049b9a4132fc978988995a61d8ea69a/cognite/extractorutils/unstable/core/base.py#L254-L261) we don't set delay at all. So technically it is set to False. But yea, in such cases, the ideal approach is to remove delay as an option or conditionally do the `filename.touch`

nithinb (approved):
LGTM. I had a comment but it does not block the PR

🐴

nithinb:
@trygveu let me know if this looks good to you

trygveu:
🦄

PR #523, opened 2026-04-26
Add content-type header for zero byte files

devendra-lohar:
This PR fixes 403 Forbidden when uploading zero-byte files to AWS clusters: AWS S3 presigned URLs sign `content-type`, so the empty-file PUT must include `Content-Type` matching the file's `mime_type`. Azure Blob doesn't sign it, which is why this only occured on AWS.

Solution: forward `file_meta.mime_type` from `_upload_only_file_reference` into `_prepare_request_data_for_empty_file`, mirroring the non-empty file path.

This also solves the exisiting bug related to greenfield and uploading multipart files with content-type set, which will result in 403.

### DIFF
--- a/cognite/extractorutils/uploader/files.py
+++ b/cognite/extractorutils/uploader/files.py
@@ -440,7 +440,7 @@ class IOFileUploadQueue(AbstractUploadQueue):
 
         for url in upload_urls:
             chunks.next_chunk()
-            resp = self._httpx_client.send(self._get_file_upload_request(url, chunks, len(chunks), file_meta.mime_type))
+            resp = self._httpx_client.send(self._get_file_upload_request(url, chunks, len(chunks), mime_type=None))
             resp.raise_for_status()
 
         completed_headers = (
--- a/cognite/extractorutils/uploader/files.py
+++ b/cognite/extractorutils/uploader/files.py
@@ -513,6 +513,11 @@ class IOFileUploadQueue(AbstractUploadQueue):
             read_file: Callable[[], BinaryIO],
             file_meta: FileMetadataOrCogniteExtractorFile,
         ) -> None:
+            # Default to `application/octet-stream` so uploads succeed
+            # regardless of cluster when the caller didn't supply a mime type.
+            if file_meta.mime_type is None:
+                file_meta.mime_type = "application/octet-stream"
+
             with read_file() as file:
                 size = super_len(file)
                 if size == 0:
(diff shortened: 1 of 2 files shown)


Hmnt39 (dismissed):
LGTM

cognite/extractorutils/uploader/files.py
```diff
@@ -417,10 +417,13 @@ def _prepare_request_data_for_empty_file(self, url_str: str) -> Request:
...
         )
 
+        if mime_type is not None:
+            headers.update({"Content-Type": mime_type})
+
         return Request(method="PUT", url=upload_url, headers=headers, content=EMPTY_CONTENT)
 
     def _upload_only_file_reference(self, file_meta: FileMetadataOrCogniteExtractorFile, url_str: str) -> None:
-        request_data = self._prepare_request_data_for_empty_file(url_str)
+        request_data = self._prepare_request_data_for_empty_file(url_str, file_meta.mime_type)
```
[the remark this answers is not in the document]
devendra-lohar: functionally it makes no difference as CDF gives back exactly the mimeType we sent, so file_meta.mime_type and file_meta_response.mime_type are always equal at that point.

Hmnt39 (approved):

Hmnt39 (changes requested):
Test cases are failing

cognite/extractorutils/uploader/files.py
```diff
@@ -437,7 +440,7 @@ def _upload_multipart(self, size: int, file: BinaryIO, file_meta: FileMetadataOr
 
         for url in upload_urls:
             chunks.next_chunk()
-            resp = self._httpx_client.send(self._get_file_upload_request(url, chunks, len(chunks), file_meta.mime_type))
+            resp = self._httpx_client.send(self._get_file_upload_request(url, chunks, len(chunks), mime_type=None))
```
Hmnt39, line 445: Maybe add a comment here explaining why mime_type=None is required, for future reference.

Hmnt39 (changes requested):
nit : There are duplicated code in the test cases, can we create a function and use that?

cognite/extractorutils/uploader/files.py
devendra-lohar: added

Hmnt39 (approved):

sighol (approved):
:unicorn: -waiting-for-risk-review +waiting-for-team +risk-review-ongoing +assignee:@sighol

PR #525, opened 2026-04-29
DOG-6871 Fix incorrect config version associated with invalid config error

Yaseen-A-Khan:
This PR adds:

**Fix:** 
On remote config validation failure, unstable Runtime now sends /odin/checkin errors with type: "config" and configRevision from InvalidConfigError.attempted_revision, so Odin shows the correct config version.

**Changes:** 
Optional type / config_revision on unstable Error DTO; populate them in `_safe_get_application_config`.

And updated the changelog and versions

### DIFF
--- a/cognite/extractorutils/unstable/core/runtime.py
+++ b/cognite/extractorutils/unstable/core/runtime.py
@@ -350,6 +350,8 @@ class Runtime(Generic[ExtractorType]):
                     description=error_message,
                     details=None,
                     task=None,
+                    type="config" if isinstance(e, InvalidConfigError) and e.attempted_revision is not None else None,
+                    config_revision=e.attempted_revision if isinstance(e, InvalidConfigError) else None,
                 )
 
                 self._cognite_client.post(
--- a/cognite/extractorutils/unstable/core/runtime.py
+++ b/cognite/extractorutils/unstable/core/runtime.py
@@ -342,8 +342,8 @@ class Runtime(Generic[ExtractorType]):
                 prev_error = error_message
 
                 ts = now()
-                is_config_error = isinstance(e, InvalidConfigError)
-                revision = e.attempted_revision if is_config_error else None
+                config_exc = e if isinstance(e, InvalidConfigError) else None
+                revision = config_exc.attempted_revision if config_exc is not None else None
                 error = Error(
                     external_id=str(uuid4()),
                     level=ErrorLevel.fatal,
@@ -352,7 +352,7 @@ class Runtime(Generic[ExtractorType]):
                     description=error_message,
                     details=None,
                     task=None,
-                    type="config" if is_config_error else None,
+                    type="config" if config_exc is not None else None,
                     config_revision=revision,
                 )
 
--- a/cognite/extractorutils/unstable/core/runtime.py
+++ b/cognite/extractorutils/unstable/core/runtime.py
@@ -358,7 +358,7 @@ class Runtime(Generic[ExtractorType]):
                     f"/api/v1/projects/{self._cognite_client.config.project}/odin/checkin",
                     json={
                         "externalId": connection_config.integration.external_id,
-                        "errors": [error.model_dump(mode="json")],
+                        "errors": [error.model_dump(mode="json", by_alias=True)],
                     },
                     headers={"cdf-version": "alpha"},
                 )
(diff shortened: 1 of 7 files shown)


tests/test_unstable/test_runtime.py
```diff
@@ -158,6 +158,46 @@ def cancel_after_delay() -> None:
...
+    )
+
+    errors = cognite_client.get(
+        url=f"/api/v1/projects/{cognite_client.config.project}/integrations/errors",
+        params={"integration": connection_config.integration.external_id},
+        headers={"cdf-version": "alpha"},
+    ).json()
+
+    assert len(errors["items"]) >= 1
+    assert errors["items"][0].get("activeConfigRevision") == 1
```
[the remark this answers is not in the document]
Yaseen-A-Khan: /gemini is this addressed now?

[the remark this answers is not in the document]
Yaseen-A-Khan: /gemini why are tests failing? what is the main reason behind the failure ?

tests/test_unstable/test_runtime.py
```diff
@@ -158,6 +158,73 @@ def cancel_after_delay() -> None:
...
+            "activeConfigRevision": expected_revision,  # Match the failing config revision
+        },
+        headers={"cdf-version": "alpha"},
+    )
+
+    runtime = Runtime(TestExtractor)
+    runtime._cognite_client = cognite_client
+    runtime.RETRY_CONFIG_INTERVAL = 1
+
+    def cancel_after_delay() -> None:
```
Hmnt39, line 207: Do we need a delay here or we can just call `_safe_get_application_config` and cancel the cancellation_token immediately?

cognite/extractorutils/unstable/core/runtime.py
```diff
@@ -350,13 +351,15 @@ def _safe_get_application_config(
                     description=error_message,
                     details=None,
                     task=None,
+                    type="config" if revision is not None else None,
```
Hmnt39: Should all InvalidConfigError be tagged as "config" regardless of revision? If yes, then mapping error to InvalidConfigError will be good?

```suggestion
is_config_error = isinstance(e, InvalidConfigError)
revision = e.attempted_revision if is_config_error else None

type = "config" if is_config_error else None
config_revision = revision
```

Hmnt39 (changes requested):
Minor nit, overall looks good

tests/test_unstable/test_runtime.py
Yaseen-A-Khan: `_safe_get_application_config` keeps trying until something cancels it, so it never finishes on its own when config is incorrect.
If cancel happens too soon the first error/check-in might not be posted yet.
so we add a short sleep to make that order reliable.

cognite/extractorutils/unstable/core/runtime.py
Yaseen-A-Khan: applying the above suggestion, mypy checks failed but updated the config error check accordingly.

cognite/extractorutils/unstable/core/runtime.py
```diff
@@ -350,13 +352,15 @@ def _safe_get_application_config(
                     description=error_message,
                     details=None,
                     task=None,
+                    type="config" if config_exc is not None else None,
+                    config_revision=revision,
```
devendra-lohar, line 356: When type is None, Odin will defaul the type to 'general' I think. Worth checking how Odin handles if type is sent None (default to general) and config_revision is also sent None. Since I think for general type Error object does not have config_revision at all.

Yaseen-A-Khan: I agree with you, it does default to "general" 
But the odin never sees `"type": null` because `CogniteModel.model_dump` always applies `exclude_none=True` so None fields are dropped entirely from the JSON, It just sees a bare error with no type field

devendra-lohar: So we never send the fields which are null?

Yaseen-A-Khan: Yes, type and configRevision are completely omitted when sending to odin

devendra-lohar (approved):
LGTM.

thorkildcognite:
@Hmnt39 Are your objections fully solved with the newest changes? (if so, can you remove your "Change Requested"?)

Hmnt39 (approved):

psalaberria002:
🦄

PR #528, opened 2026-05-07
fix mypy check failure

Yaseen-A-Khan:
This PR adds:

- a fix to mypy check error emerged with update to the mypy version `1.20.2` --> `2.0.0`


devendra-lohar (approved):

nithinb (approved):
risk review ok

Had an online discussion about using the TypeVar but since python only allows it in the context of methods / classes it wouldn't help here.

PR #532, opened 2026-05-19
feat(odin): Add custom action support

vikramlc-cognite:
### PR short summary:
- Adds support for custom actions - named operations that extractor authors can define and expose to be triggered at runtime by an operator.

---
### What changed:
- New actions.py with CustomAction (holds name + callable), ActionContext (passed to the callable, handles error logging), and ActionTarget (type alias for the callable).
- Unit tests covering the new classes.

---
### Why it changed:
- Extractors only supported start/stop of tasks. Odin needs a way to trigger ad-hoc operations like, say fetch logs, without going through the task lifecycle.

---
### What to focus on:
- ActionContext._new_error - errors default to using the action's name as task_name unless overridden.
- Logger naming format: {extractor_id}.action.{action_name} - confirm this matches what Odin expects.

---
### Test evidence:
- All unit tests under tests/test_unstable/test_actions.py pass.

---
### Risks or unknowns:
- No wiring to the Extractor base class yet so no risk yet.

### DIFF
--- a/cognite/extractorutils/unstable/core/actions.py
+++ b/cognite/extractorutils/unstable/core/actions.py
@@ -50,6 +50,7 @@ class ActionContext(CogniteLogger):
             level=level,
             description=description,
             details=details,
+            task_name=task_name or self._action.name,
         )
 
 
--- a/cognite/extractorutils/unstable/core/actions.py
+++ b/cognite/extractorutils/unstable/core/actions.py
@@ -20,6 +20,9 @@ class ActionContext(CogniteLogger):
     Context for a custom action invocation.
 
     This class is used to log errors and messages related to the action invocation.
+
+    ``external_id`` and ``call_metadata`` come from the pending action payload sent by Odin and are
+    available for use when reporting results back to the server.
     """
 
     def __init__(
--- a/cognite/extractorutils/unstable/core/actions.py
+++ b/cognite/extractorutils/unstable/core/actions.py
@@ -6,7 +6,6 @@ import logging
 from collections.abc import Callable
 from typing import TYPE_CHECKING
 
-from cognite.extractorutils.unstable.core._dto import JSONType
 from cognite.extractorutils.unstable.core.errors import Error, ErrorLevel
 from cognite.extractorutils.unstable.core.logger import CogniteLogger
 
@@ -28,7 +27,7 @@ class ActionContext(CogniteLogger):
         action: "CustomAction",
         extractor: "Extractor",
         external_id: str,
-        call_metadata: JSONType | None = None,
+        call_metadata: dict[str, str] | None = None,
     ) -> None:
         super().__init__()
         self._action = action
@@ -50,7 +49,7 @@ class ActionContext(CogniteLogger):
             level=level,
             description=description,
             details=details,
-            task_name=task_name or self._action.name,
+            task_name=task_name if task_name is not None else self._action.name,
         )
 
 
--- a/cognite/extractorutils/unstable/core/actions.py
+++ b/cognite/extractorutils/unstable/core/actions.py
@@ -36,9 +36,7 @@ class ActionContext(CogniteLogger):
         self.external_id = external_id
         self.call_metadata = call_metadata
 
-        self._logger = logging.getLogger(
-            f"{self._extractor.EXTERNAL_ID}.action.{self._action.name.replace(' ', '')}"
-        )
+        self._logger = logging.getLogger(f"{self._extractor.EXTERNAL_ID}.action.{self._action.name.replace(' ', '')}")
 
     def _new_error(
         self,
--- a/cognite/extractorutils/unstable/core/actions.py
+++ b/cognite/extractorutils/unstable/core/actions.py
@@ -1,0 +1,80 @@
+"""
+This module defines the base classes for custom actions in the extractor framework.
+"""
+
+import logging
+from collections.abc import Callable
+from typing import TYPE_CHECKING
+
+from cognite.extractorutils.unstable.core._dto import JSONType
+from cognite.extractorutils.unstable.core.errors import Error, ErrorLevel
+from cognite.extractorutils.unstable.core.logger import CogniteLogger
+
+if TYPE_CHECKING:
+    from cognite.extractorutils.unstable.core.base import Extractor
+
+__all__ = ["ActionContext", "ActionTarget", "CustomAction"]
+
+
+class ActionContext(CogniteLogger):
+    """
+    Context for a custom action invocation.
+
+    This class is used to log errors and messages related to the action invocation.
+    """
+
+    def __init__(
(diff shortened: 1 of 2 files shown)


cognite/extractorutils/unstable/core/actions.py
```diff
@@ -0,0 +1,79 @@
...
+
+    This class is used to log errors and messages related to the action invocation.
+    """
+
+    def __init__(
+        self,
+        action: "CustomAction",
+        extractor: "Extractor",
+        external_id: str,
+        call_metadata: JSONType | None = None,
```
Yaseen-A-Khan: accepts `call_metadata: JSONType | None` but `Action.call_metadata` in the DTO is typed as `dict[str, str] | None` 
is there a possibility that this might cause a problem or is there a validation that shows that this wont happen?

tests/test_unstable/test_actions.py
```diff
@@ -0,0 +1,107 @@
...
+
+    assert action.description is None
+
+
+def test_action_context_attributes(mock_extractor: MagicMock, simple_action: CustomAction) -> None:
+    ctx = ActionContext(
+        action=simple_action,
+        extractor=mock_extractor,
+        external_id="triggered-action-ext-id",
+        call_metadata={"key": "value"},
```
Yaseen-A-Khan, line 50: if there is a possibility that other types can also be taken then there should be some type validation test for that

cognite/extractorutils/unstable/core/actions.py
```diff
@@ -0,0 +1,79 @@
...
+        description: str,
+        *,
+        details: str | None = None,
+        task_name: str | None = None,
+    ) -> Error:
+        return self._extractor._new_error(
+            level=level,
+            description=description,
+            details=details,
+            task_name=task_name or self._action.name,
```
Yaseen-A-Khan: task_name=task_name or self._action.name uses truthiness, so an explicit empty string "" would still default to self._action.name. Is it intended?

tests/test_unstable/test_actions.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def test_action_context_logger_name_strips_spaces(mock_extractor: MagicMock) -> None:
+    action = CustomAction(name="process data", target=lambda ctx: None)
+    ctx = ActionContext(action=action, extractor=mock_extractor, external_id="ext-id")
+
+    assert ctx._logger.name == "test-extractor.action.processdata"
+
+
+def test_action_context_error_delegates_to_extractor(mock_extractor: MagicMock, simple_action: CustomAction) -> None:
```
Yaseen-A-Khan: can we parametrize `test_action_context_error_delegates_to_extractor` and `test_action_context_error_passes_explicit_task_name` as both test the _new_error with different parameters

tests/test_unstable/test_actions.py
```diff
@@ -0,0 +1,107 @@
...
+def test_action_context_error_passes_explicit_task_name(mock_extractor: MagicMock, simple_action: CustomAction) -> None:
+    ctx = ActionContext(action=simple_action, extractor=mock_extractor, external_id="ext-id")
+
+    ctx._new_error(level=ErrorLevel.error, description="Fail", details="details", task_name="custom-task")
+
+    call_kwargs = mock_extractor._new_error.call_args.kwargs
+    assert call_kwargs["task_name"] == "custom-task"
+
+
+def test_action_target_is_callable() -> None:
```
Yaseen-A-Khan: can we also test by actually calling ?

Yaseen-A-Khan (commented):
I have few comments, maybe you can help clarifying them.

tests/test_unstable/test_actions.py
vikramlc-cognite: Now that the type is correctly narrowed to dict[str, str] | None, this doesn't need runtime validation tests. Type safety here is enforced by the type checker (mypy/pyright), not at runtime.

cognite/extractorutils/unstable/core/actions.py
vikramlc-cognite: Fixed.

tests/test_unstable/test_actions.py
vikramlc-cognite: Done.

tests/test_unstable/test_actions.py
vikramlc-cognite: Done.

cognite/extractorutils/unstable/core/actions.py
vikramlc-cognite: Idea was to reuse JSONType in utils but you are correct here. We can make it same as in Odin.

cognite/extractorutils/unstable/core/actions.py
```diff
@@ -0,0 +1,78 @@
...
+        action: "CustomAction",
+        extractor: "Extractor",
+        external_id: str,
+        call_metadata: dict[str, str] | None = None,
+    ) -> None:
+        super().__init__()
+        self._action = action
+        self._extractor = extractor
+        self.external_id = external_id
+        self.call_metadata = call_metadata
```
Yaseen-A-Khan, line 39: Constructor sets self.external_id and self.call_metadata from the server Action DTO (future). No methods read them in this PR.

tests/test_unstable/test_actions.py
```diff
@@ -0,0 +1,109 @@
...
+
+@pytest.mark.parametrize(
+    "task_name,expected_task_name",
+    [
+        (None, "my action"),
+        ("custom-task", "custom-task"),
+        ("", ""),
+    ],
+)
+def test_action_context_error_task_name(
```
Yaseen-A-Khan, line 86: Parametrize case ( "", "" ) explicit empty string is forwarded, not replaced with action name.
only None triggers default to self._action.name. Empty string may produce odd CDF error reports if used by mistake.

tests/test_unstable/test_actions.py
```diff
@@ -0,0 +1,109 @@
...
+    mock_extractor: MagicMock,
+    simple_action: CustomAction,
+    task_name: str | None,
+    expected_task_name: str,
+) -> None:
+    ctx = ActionContext(action=simple_action, extractor=mock_extractor, external_id="ext-id")
+
+    ctx._new_error(level=ErrorLevel.warning, description="Something went wrong", task_name=task_name)
+
+    call_kwargs = mock_extractor._new_error.call_args.kwargs
```
Yaseen-A-Khan, line 98: Tests only assert task_name in call_kwargs. details and level are not asserted.

Yaseen-A-Khan (changes requested):

Yaseen-A-Khan (approved):
LGTM

psalaberria002 (approved):
🦄

PR #534, opened 2026-05-25
feat(odin): Add cancellation token to TaskContext for cancel task

vikramlc-cognite:
### PR Short Summary
- Adds a per-task cancellation_token (CancellationToken) to TaskContext so individual tasks can be cancelled
  cooperatively without stopping the whole extractor. Each task receives a child token derived from the extractor-level
   token, preserving parent-propagation semantics.

### What changed
- Added cancellation_token: CancellationToken as a required constructor parameter and public attribute on TaskContext.
- Updated all three TaskContext(...) call sites (scheduled task lambda, startup task thread pool, continuous task thread) to pass self.cancellation_token.create_child_token().
- Added tests.

### Why it changed
- Prerequisite for implementing **stop_task** actions: cancelling a single running batch task requires each task to own a cancellation token that can be signalled independently. Previously, task code could only check **extractor.cancellation_token**, which cancels the entire extractor process.

- Check if the cancellation token passed to TaskContext is correct.
- Will the cancellation token only cancel the child token and not parent token when it is cancelled.

### Test evidence
- All newly added tests pass.

### Risks and unknowns
- None. Purely additive

### Rollout and rollback
- Reverting is a straight revert of the 4 changed files.

### DIFF
--- a/tests/test_unstable/test_task_context.py
+++ b/tests/test_unstable/test_task_context.py
@@ -1,0 +1,45 @@
+from unittest.mock import MagicMock
+
+from cognite.extractorutils.threading import CancellationToken
+from cognite.extractorutils.unstable.core.tasks import StartupTask, TaskContext
+
+
+def _make_context(parent_token: CancellationToken) -> tuple[TaskContext, CancellationToken]:
+    task = StartupTask(name="test task", target=lambda ctx: None)
+    extractor = MagicMock()
+    extractor.EXTERNAL_ID = "test-extractor"
+    child_token = parent_token.create_child_token()
+    ctx = TaskContext(task=task, extractor=extractor, cancellation_token=child_token)
+    return ctx, child_token
+
+
+def test_cancellation_token_not_cancelled_initially() -> None:
+    parent = CancellationToken()
+    ctx, _ = _make_context(parent)
+    assert not ctx.cancellation_token.is_cancelled
+
+
+def test_cancelling_task_token_does_not_affect_extractor_token() -> None:
+    parent = CancellationToken()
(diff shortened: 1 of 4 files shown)


Toshad (commented):
Can you add a test with dummy tasks verifying that cancelling parent token cancels child task?

Edit: Nvm, it's already there

tests/test_unstable/test_task_context.py
```diff
@@ -0,0 +1,45 @@
...
+
+from cognite.extractorutils.threading import CancellationToken
+from cognite.extractorutils.unstable.core.tasks import StartupTask, TaskContext
+
+
+def _make_context(parent_token: CancellationToken) -> tuple[TaskContext, CancellationToken]:
+    task = StartupTask(name="test task", target=lambda ctx: None)
+    extractor = MagicMock()
+    extractor.EXTERNAL_ID = "test-extractor"
+    child_token = parent_token.create_child_token()
```
Toshad: You shouldn't explicitly create child token here, you should add task and then test against that.
These tests end up testing the create_child_token() function rather than the changes in this diff.

Toshad (changes requested):

tests/test_unstable/test_task_context.py
vikramlc-cognite: Good point. [Done](https://github.com/cognitedata/python-extractor-utils/pull/534/commits/7c3238788f4468e9e151c30523b1f91fa496c207).

Toshad (approved):

sighol (commented):
🦄

PR #535, opened 2026-05-27
feat(Odin): Integrate actions in checkin worker

vikramlc-cognite:
### PR Short Summary
- Wires up action dispatch in CheckinWorker: every checkin and startup response can now carry pending_actions, which are handed off to a configurable dispatcher callback, and any action outcome reports (action_updates) are carried back in the next outgoing checkin request.

---
### Type of change
- [ ] Bug fix
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / tooling / CI

---
### What changed
cognite/extractorutils/unstable/core/checkin_worker.py
- New state: _action_updates: list[ActionUpdate], _action_dispatcher: Callable | None
- set_action_dispatcher(dispatcher) — registers the callback; calling it twice replaces the previous one
- queue_action_update(update) — thread-safe append under _lock; intended to be called from action handler threads
- _handle_checkin_response — fires _action_dispatcher on a named daemon Thread("ActionDispatcher") when pending_actions is present, so action execution never blocks the checkin interval
- try_write_checkin — drains _action_updates under _lock before each POST, passes them as action_updates (omitted when empty via exclude_none)
- _requeue_checkin — gained action_updates parameter; prepends (not appends) failed-batch updates back to _action_updates to preserve send order across retries
  
tests/test_unstable/test_checkin_worker_actions.py (new file)
- 8 tests covering all acceptance criteria, following the same requests_mock + gzip-decoded body pattern as test_checkin_worker.py

---
### Why it changed
- Implements EDGE-182: actions integration — checkin worker needs to parse pending_actions from Odin responses, dispatch them to extractor logic, and carry back outcome reports.
- Odin spec: every checkin/startup response can include pending actions; the worker is the sole interface to Odin, so dispatch must live here

---
- _requeue_checkin uses prepend, not append ((action_updates or []) + self._action_updates). This preserves ordering across retries: updates from a failed batch must precede updates that arrived after the failure. The test test_action_updates_requeue_order_preserved_across_retry validates this.
- Daemon thread for dispatch — the ActionDispatcher thread is daemon so it doesn't block extractor shutdown. The trade-off is that in-flight action handling is dropped on process exit without a clean shutdown window.
- Lock scope in try_write_checkin — the drain (_action_updates[:] + .clear()) is a single critical section. A queue_action_update racing the drain either lands in the current batch (wins the lock first) or the next one (loses it) — never lost.
- extra="ignore" on Action and CheckinResponse — deliberate: Odin may add fields before the SDK is updated.

  ---
### Test evidence
- All tests pass.

---
### Risks and unknowns
- If an action handler calls queue_action_update after the drain in try_write_checkin but before the POST succeeds or fails, that update is not part of the current batch and will go out on the next checkin — this is correct but means outcome reporting has at most one checkin interval of latency.
- No backpressure on _action_updates: if action handlers produce updates faster than checkins can flush them, the list grows unbounded. Not a concern at current action volumes but worth noting.

---
### Rollout and rollback
- No migration, no feature flag. _action_dispatcher defaults to None, so all existing extractors that don't call set_action_dispatcher are completely unaffected — the new code paths are dead without a dispatcher registered.

---
### Checklist
- [x] Self-reviewed the diff
- [x] Tests added (test_checkin_worker_actions.py — 8 tests)
- [x] Docs: N/A (internal API, no public docs affected)
- [x] No breaking changes — all new fields on existing models are | None with defaults

### DIFF
--- a/cognite/extractorutils/unstable/core/checkin_worker.py
+++ b/cognite/extractorutils/unstable/core/checkin_worker.py
@@ -40,6 +40,7 @@ from cognite.extractorutils.util import now
 
 DEFAULT_SLEEP_INTERVAL = STARTUP_BACKOFF_SECONDS = 30.0
 MAX_ERRORS_PER_CHECKIN = MAX_TASK_UPDATES_PER_CHECKIN = 1000
+MAX_ACTION_UPDATES_PER_CHECKIN = 100
 rng = SystemRandom()
 
 
@@ -130,6 +131,9 @@ class CheckinWorker:
             dispatcher (Callable[[list[Action]], None]): A callback invoked on a daemon thread
                                                          when the server returns pending_actions.
         """
+        # No lock needed: a single reference store is atomic under CPython's GIL.
+        # _handle_checkin_response snapshots self._action_dispatcher into a local variable
+        # before use, so a concurrent replacement is safe.
         self._action_dispatcher = dispatcher
 
     def queue_action_update(self, update: ActionUpdate) -> None:
@@ -301,6 +305,8 @@ class CheckinWorker:
                 for error in new_errors:
                     del self._errors[error.external_id]
                 task_updates: list[TaskUpdate] = []
+                # action_updates are not suppressed here: outcome reports should reach
+                # the server as soon as possible regardless of startup state.
             else:
                 new_errors = list(self._errors.values())
                 self._errors.clear()
@@ -376,8 +382,12 @@ class CheckinWorker:
         task_updates(list[TaskUpdate]): The task updates to write.
         """
         with self._lock:
-            action_updates = self._action_updates
-            self._action_updates = []
+            if len(self._action_updates) > MAX_ACTION_UPDATES_PER_CHECKIN:
+                action_updates = self._action_updates[:MAX_ACTION_UPDATES_PER_CHECKIN]
+                self._action_updates = self._action_updates[MAX_ACTION_UPDATES_PER_CHECKIN:]
+            else:
+                action_updates = self._action_updates
+                self._action_updates = []
 
         checkin_request = CheckinRequest(
             external_id=self._integration,
--- a/cognite/extractorutils/unstable/core/checkin_worker.py
+++ b/cognite/extractorutils/unstable/core/checkin_worker.py
@@ -43,6 +43,17 @@ MAX_ERRORS_PER_CHECKIN = MAX_TASK_UPDATES_PER_CHECKIN = 1000
 rng = SystemRandom()
 
 
+def _safe_dispatch(
+    dispatcher: Callable[[list[Action]], None],
+    actions: list[Action],
+    logger: Logger,
+) -> None:
+    try:
+        dispatcher(actions)
+    except Exception:
+        logger.exception("Unhandled exception in action dispatcher")
+
+
 class CheckinWorker:
     """
     A worker to manage how we report check-ins to the Integrations API.
@@ -52,9 +63,6 @@ class CheckinWorker:
     1. Manage how we batch errors and task updates.
     2. Manage how we handle retries and backoff.
     """
-
-    _lock = RLock()
-    _flush_lock = RLock()
 
     def __init__(
         self,
@@ -89,6 +97,8 @@ class CheckinWorker:
         self._task_updates: list[TaskUpdate] = []
         self._action_updates: list[ActionUpdate] = []
         self._action_dispatcher: Callable[[list[Action]], None] | None = None
+        self._lock = RLock()
+        self._flush_lock = RLock()
 
     @property
     def active_revision(self) -> ConfigRevision:
@@ -243,16 +253,16 @@ class CheckinWorker:
                 self._active_revision = checkin_response.last_config_revision
 
         if checkin_response.pending_actions and self._action_dispatcher is not None:
-            dispatcher = self._action_dispatcher
-            actions = checkin_response.pending_actions
-
-            def _safe_dispatch() -> None:
-                try:
-                    dispatcher(actions)
-                except Exception:
-                    self._logger.exception("Unhandled exception in action dispatcher")
-
-            Thread(target=_safe_dispatch, name="ActionDispatcher", daemon=True).start()
+            # Each checkin tick may spawn its own ActionDispatcher thread. Overlapping
+            # dispatches are intentional: the checkin interval is much longer than a
+            # typical dispatch, and actions carry server-assigned external IDs so
+            # duplicate delivery is handled by the caller.
+            Thread(
+                target=_safe_dispatch,
+                args=(self._action_dispatcher, checkin_response.pending_actions, self._logger),
+                name="ActionDispatcher",
+                daemon=True,
+            ).start()
 
     def flush(self, cancellation_token: CancellationToken) -> None:
         """
@@ -263,9 +273,10 @@ class CheckinWorker:
         """
         with self._flush_lock:
             self._logger.debug(
-                "Going to report check-in with %d errors and %d task updates.",
+                "Going to report check-in with %d errors, %d task updates and %d action updates.",
                 len(self._errors),
                 len(self._task_updates),
+                len(self._action_updates),
             )
             self.report_checkin(cancellation_token)
 
--- a/cognite/extractorutils/unstable/core/checkin_worker.py
+++ b/cognite/extractorutils/unstable/core/checkin_worker.py
@@ -252,14 +252,15 @@ class CheckinWorker:
                     self._on_revision_change(checkin_response.last_config_revision)
                 self._active_revision = checkin_response.last_config_revision
 
-        if checkin_response.pending_actions and self._action_dispatcher is not None:
+        dispatcher = self._action_dispatcher
+        if checkin_response.pending_actions and dispatcher is not None:
             # Each checkin tick may spawn its own ActionDispatcher thread. Overlapping
             # dispatches are intentional: the checkin interval is much longer than a
             # typical dispatch, and actions carry server-assigned external IDs so
             # duplicate delivery is handled by the caller.
             Thread(
                 target=_safe_dispatch,
-                args=(self._action_dispatcher, checkin_response.pending_actions, self._logger),
+                args=(dispatcher, checkin_response.pending_actions, self._logger),
                 name="ActionDispatcher",
                 daemon=True,
             ).start()
@@ -375,8 +376,8 @@ class CheckinWorker:
         task_updates(list[TaskUpdate]): The task updates to write.
         """
         with self._lock:
-            action_updates = self._action_updates[:]
-            self._action_updates.clear()
+            action_updates = self._action_updates
+            self._action_updates = []
 
         checkin_request = CheckinRequest(
             external_id=self._integration,
@@ -454,7 +455,8 @@ class CheckinWorker:
                 if error.external_id not in self._errors:
                     self._errors[error.external_id] = error
             self._task_updates.extend(task_updates or [])
-            self._action_updates = (action_updates or []) + self._action_updates
+            if action_updates:
+                self._action_updates = action_updates + self._action_updates
 
     def _wrap_checkin_like_request(self, request: Callable[[], Response]) -> bool:
         exception_to_report: Exception | None = None
--- a/cognite/extractorutils/unstable/core/checkin_worker.py
+++ b/cognite/extractorutils/unstable/core/checkin_worker.py
@@ -243,12 +243,16 @@ class CheckinWorker:
                 self._active_revision = checkin_response.last_config_revision
 
         if checkin_response.pending_actions and self._action_dispatcher is not None:
-            Thread(
-                target=self._action_dispatcher,
-                args=(checkin_response.pending_actions,),
-                name="ActionDispatcher",
-                daemon=True,
-            ).start()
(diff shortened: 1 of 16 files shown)


cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -325,10 +375,15 @@ def try_write_checkin(self, errors: list[Error], task_updates: list[TaskUpdate])
...
         """
+        with self._lock:
+            action_updates = self._action_updates
+            self._action_updates = []
+
         checkin_request = CheckinRequest(
             external_id=self._integration,
             errors=list(map(DtoError.from_internal, errors)) if len(errors) > 0 else None,
             task_events=task_updates if len(task_updates) > 0 else None,
+            action_updates=action_updates or None,
```
Yaseen-A-Khan, line 385: ActionUpdateList max is 100, but we drain the whole queue into one CheckinRequest. At 101+ updates, Pydantic raises ValidationError outside _wrap_checkin_like_request — that kills the checkin thread.

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,412 @@
...
+
+@pytest.fixture
+def mock_checkin_with_actions(
+    connection_config: ConnectionConfig,
+    checkin_bag: list,
+    action_updates_bag: list,
+) -> Callable[[requests_mock_lib.Mocker, list[dict] | None, int], None]:
+    """
+    Like mock_checkin_request, but also captures actionUpdates and can embed
+    pendingActions in the response JSON.
```
Yaseen-A-Khan: PR says startup carries pendingActions, and the code path exists, but no test mocks startup with pendingActions

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,412 @@
...
+    connection_config: ConnectionConfig,
+) -> None:
+    """No exception is raised when pending_actions arrive but no dispatcher is registered.
+
+    Extractors that don't use action dispatch must not crash when the server sends actions.
+    """
+    worker = _make_worker(connection_config)
+    assert worker._action_dispatcher is None
+
+    worker._handle_checkin_response(
```
Yaseen-A-Khan: some tests use `requests_mock.real_http = True` and some use `_handle_checkin_response` from mock directly, is this intended?

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,412 @@
...
+    worker = _make_worker(connection_config)
+    updates = [ActionUpdate(external_id=f"act-{i}", status=ActionStatus.succeeded) for i in range(50)]
+
+    threads = [Thread(target=worker.queue_action_update, args=(u,)) for u in updates]
+    for t in threads:
+        t.start()
+    for t in threads:
+        t.join()
+
+    assert len(worker._action_updates) == 50
```
Yaseen-A-Khan, line 130: can we also assert on the external_id, maybe that would be stronger as well.

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,412 @@
...
+import logging
+import threading
+from collections.abc import Callable, Generator
+from datetime import datetime, timezone
+from multiprocessing import Event, Queue
+from threading import Thread
+from typing import Any
+
+import pytest
+import requests_mock as requests_mock_lib
```
Yaseen-A-Khan: I think we can use this just as `import requests_mock` to maintain consistency

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,412 @@
...
+) -> None:
+    """Calling set_action_dispatcher a second time replaces the first dispatcher."""
+    worker = _make_worker(connection_config)
+    first = lambda actions: None  # noqa: E731
+    second = lambda actions: None  # noqa: E731
+
+    worker.set_action_dispatcher(first)
+    worker.set_action_dispatcher(second)
+
+    assert worker._action_dispatcher is second
```
Yaseen-A-Khan: Prefer behavioral assertion (first never called, second is).

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,412 @@
...
+
+        def emit(self, record: logging.LogRecord) -> None:
+            self.records.append(record)
+            self.received.set()
+
+    handler = _CapturingHandler()
+    worker._logger.addHandler(handler)
+    worker._logger.setLevel(logging.ERROR)
+
+    def bad_dispatcher(actions: list) -> None:
```
Yaseen-A-Khan: `bad_dispatcher(actions: list)` drops list[Action] type hint

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,412 @@
...
+        }
+    )
+
+    assert fired.wait(timeout=2), "Dispatcher was never called"
+    assert len(received) == 2
+    assert received[0].external_id == "act-1"
+    assert received[1].external_id == "act-2"
+
+
+def test_dispatcher_exception_is_logged_not_silently_swallowed(
```
Yaseen-A-Khan, line 250: can we add stronger assertions for this?

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,412 @@
...
+"""
+
+import gzip
+import json
+import logging
+import threading
+from collections.abc import Callable, Generator
+from datetime import datetime, timezone
+from multiprocessing import Event, Queue
+from threading import Thread
```
Yaseen-A-Khan: duplicate import of threading

Yaseen-A-Khan (changes requested):
I have some comments including changes and nits
and also can you make sure some docstrings are updated, some are incomplete

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -225,9 +274,10 @@ def flush(self, cancellation_token: CancellationToken) -> None:
         """
         with self._flush_lock:
             self._logger.debug(
-                "Going to report check-in with %d errors and %d task updates.",
+                "Going to report check-in with %d errors, %d task updates and %d action updates.",
                 len(self._errors),
                 len(self._task_updates),
+                len(self._action_updates),
             )
             self.report_checkin(cancellation_token)
```
nithinb, line 276: I can't comment below, but in `report_checkin` you explicitly set `task_updates: list[TaskUpdate] = []` but `_action_updates` are not touched. Is that expected?

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -108,6 +122,28 @@ def reset_startup(self) -> None:
...
 
+    def set_action_dispatcher(self, dispatcher: Callable[[list[Action]], None]) -> None:
+        """
+        Set the dispatcher for handling pending actions from checkin responses.
+
+        Arguments:
+            dispatcher (Callable[[list[Action]], None]): A callback invoked on a daemon thread
+                                                         when the server returns pending_actions.
+        """
+        self._action_dispatcher = dispatcher
```
nithinb, line 129: I think it is worth documenting (maybe as a comment) why `self._lock` is not required here?

nithinb (changes requested):
Also one last thing is that _has_reported_startup=False has no test coverage. I think that is a case worth considering as part of the tests that you have

cognite/extractorutils/unstable/core/checkin_worker.py
vikramlc-cognite: Added inline comment: 
> action_updates are not suppressed here: outcome reports should reach odin as soon as possible regardless of startup state.

Unlike task_updates (which require startup context), action outcomes carry task IDs and are safe to send immediately.

cognite/extractorutils/unstable/core/checkin_worker.py
vikramlc-cognite: Added a comment explaining the reasoning: single reference store is atomic under CPython's GIL, and **_handle_checkin_response** already takes a local snapshot (dispatcher = self._action_dispatcher) so concurrent replacement is safe regardless.

tests/test_unstable/test_checkin_worker_actions.py
vikramlc-cognite: Added a note to the module docstring explaining the two levels: direct _handle_checkin_response calls are unit-level (dispatch logic in isolation), requests_mock path is integration-level (full HTTP cycle).

tests/test_unstable/test_checkin_worker_actions.py
```diff
@@ -0,0 +1,430 @@
...
+
+    worker.report_task_start("some-task")
+    worker.queue_action_update(ActionUpdate(external_id="act-pre", status=ActionStatus.succeeded))
+    worker.flush(cancellation_token)
+
+    assert len(checkin_bag) == 1
+    assert "actionUpdates" in checkin_bag[0]
+    assert action_updates_bag[0]["externalId"] == "act-pre"
+    assert "taskEvents" not in checkin_bag[0]
+
```
nithinb, line 355: Here, there should be validation of the actual error in checkin_bag. so assertions for actionUpdates and errors in checkin_bag.

Something along the lines of:

```python
assert "errors" in checkin_bag[0]
assert checkin_bag[0]["errors"][0]["description"] == <"Expected error message in description">

```

The rest LGTM

nithinb (changes requested):
The PR looks good overall. 

Only one assertion is something that would be good to have.

tests/test_unstable/test_checkin_worker_actions.py
vikramlc-cognite: [Added](https://github.com/cognitedata/python-extractor-utils/pull/535/commits/874c9ff9917fb97e8fc7fcf848b78e9223e0c2b6)

nithinb (approved):
LGTM

Yaseen-A-Khan (approved):
LGTM

gaelenm:
[[risk_review]](https://github.com/cognitedata/infrastructure/tree/HEAD/tools/risk_review) starting risk review

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -108,6 +123,23 @@ def reset_startup(self) -> None:
...
+        Queue an action update to be included in the next checkin request.
+
+        This method is thread-safe and may be called from action handler threads.
+
+        Arguments:
+            update (ActionUpdate): The action update to queue.
+        """
+        with self._lock:
+            self._action_updates.append(update)
+
```
gaelenm, line 142: Out of curiosity, why are you using a list behind a lock here instead of just using a [queue.Queue](https://docs.python.org/3/library/queue.html)?

cognite/extractorutils/unstable/core/checkin_worker.py
```diff
@@ -216,6 +248,16 @@ def _handle_checkin_response(self, response: JSONType) -> None:
...
 
+        dispatcher = self._action_dispatcher
+        if checkin_response.pending_actions and dispatcher is not None:
+            # Overlapping dispatches are intentional: interval >> dispatch time, and IDs are server-assigned.
+            Thread(
+                target=_safe_dispatch,
+                args=(dispatcher, checkin_response.pending_actions, self._logger),
+                name="ActionDispatcher",
+                daemon=True,
+            ).start()
```
gaelenm, line 259: I don't necessarily know the specifics, but if nothing is tracking thread completing an `join`ing them, then a thread can be stopped in the middle of the main program exits.

Have you verified that whatever environment this is running in won't suddenly stop the main program, or that there are no unsafe side effects (data corruption, etc.) from a thread stopping in the middle of the job?

cognite/extractorutils/unstable/core/checkin_worker.py
vikramlc-cognite: We need to be able to put items back at the front of the queue when a checkin fails (so retried updates go out first). queue.Queue only supports adding to the back - there's no way to insert at the front without draining and rebuilding the whole thing. The list + lock lets us do that prepend cleanly.

cognite/extractorutils/unstable/core/checkin_worker.py
vikramlc-cognite: This is a documented trade-off (see **What to focus on during review** section in the PR description). We made the thread a daemon intentionally - if it weren't, a slow action handler would block the extractor from shutting down, which is worse. As for data corruption: if a handler does something unsafe on abrupt kill (e.g. a non-atomic write), that's on the handler to write safely - the framework can't protect against that regardless of whether the thread is a daemon or not.

cognite/extractorutils/unstable/core/checkin_worker.py
gaelenm: ah, right, I missed that the requeue is prepending items

gaelenm:
[[risk_review]](https://github.com/cognitedata/infrastructure/tree/HEAD/tools/risk_review) approving risk review

gaelenm (approved):
:unicorn:

PR #536, opened 2026-05-29
feat(odin): Integrate available actions in extractor

vikramlc-cognite:
### PR Short Summary
- Adds the public API for registering custom actions (add_action / __init_actions__) and updates _get_startup_request to announce the extractor's capabilities to Odin on startup via available_actions.

### Type of change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Chore / tooling / CI

### What changed
cognite/extractorutils/unstable/core/base.py
- New _custom_actions / _running_task_tokens instance attributes
- __init_actions__() hook (called after __init_tasks__()) for subclasses to register custom actions via add_action()
- _run_task_with_token() — single entry point for task execution; creates a child cancellation token, stores it in _running_task_tokens, runs the task, then cleans up in finally
- _get_startup_request() now populates available_actions: start_task + stop_task per ScheduledTask, one custom entry per registered action; None if empty

tests/test_unstable/test_action_registration.py — new test file covering all acceptance criteria

### Why it changed
-Odin needs extractors to declare available actions in the startup request so the UI can surface them.

- _run_task_with_token: verify the store → run → pop flow, especially the identity check in finally that prevents clobbering a concurrent run's token
- add_action now validates against duplicate names and conflicts with auto-generated "Start/Stop {task}" names

### Test evidence
tests/test_unstable/test_action_registration.py  13 passed
tests/test_unstable/test_task_context.py          3 passed
tests/test_unstable/test_actions.py              10 passed
ruff check — All checks passed

### Risks and unknowns
- "Start {t.name}" / "Stop {t.name}" action name format is hardcoded here — if Odin changes its convention the names will drift. Should be defined as a constant or validated against the server contract when that lands.

### Rollout and rollback
- N/A — purely additive. Extractors that don't override __init_actions__ behave identically to before.

### Checklist
- [x] Self-reviewed the diff
- [x] Tests added or updated
- [x] Docs updated (N/A — internal framework, no public docs affected)
- [x] Breaking changes called out above and communicated to affected teams (none)

### DIFF
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -516,9 +516,6 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             self._running_task_tokens[task.name] = child_token
         try:
             task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
-        except Exception:
-            self._logger.exception(f"Task '{task.name}' raised an unhandled exception")
-            raise
         finally:
             with self._running_task_tokens_lock:
                 if self._running_task_tokens.get(task.name) is child_token:
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -466,6 +466,14 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         if any(t.name == task.name for t in self._tasks):
             raise ValueError(f"Task '{task.name}' is already registered.")
 
+        if isinstance(task, ScheduledTask):
+            conflict_names = {f"Start {task.name}", f"Stop {task.name}"}
+            for action in self._custom_actions:
+                if action.name in conflict_names:
+                    raise ValueError(
+                        f"Task name '{task.name}' would conflict with existing custom action '{action.name}'."
+                    )
+
         # Store this for later, since we'll override it with the wrapped version
         target = task.target
 
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -190,6 +190,7 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         self._tasks: list[Task] = []
         self._custom_actions: list[CustomAction] = []
         self._running_task_tokens: dict[str, CancellationToken] = {}
+        self._running_task_tokens_lock = RLock()
         self._start_time: datetime
 
         self.metrics: BaseMetrics = self._load_metrics(config.metrics_class)
@@ -359,6 +360,17 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         Args:
             action: The custom action to register.
         """
+        if any(a.name == action.name for a in self._custom_actions):
+            raise ValueError(f"Custom action '{action.name}' is already registered.")
+
+        reserved_names = {
+            name for t in self._tasks if isinstance(t, ScheduledTask) for name in (f"Start {t.name}", f"Stop {t.name}")
+        }
+        if action.name in reserved_names:
+            raise ValueError(
+                f"Custom action name '{action.name}' conflicts with an auto-generated scheduled task action."
+            )
+
         self._custom_actions.append(action)
 
     def _set_runtime_message_queue(self, queue: Queue) -> None:
@@ -500,13 +512,14 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         future start_task action handler.
         """
         child_token = self.cancellation_token.create_child_token()
-        self._running_task_tokens[task.name] = child_token
+        with self._running_task_tokens_lock:
+            self._running_task_tokens[task.name] = child_token
         try:
             task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
         finally:
-            # Only remove our token; a concurrent run may have already replaced it
-            if self._running_task_tokens.get(task.name) is child_token:
-                self._running_task_tokens.pop(task.name, None)
+            with self._running_task_tokens_lock:
+                if self._running_task_tokens.get(task.name) is child_token:
+                    self._running_task_tokens.pop(task.name, None)
 
     def start(self) -> None:
         """
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -477,9 +477,8 @@ class Extractor(Generic[ConfigType], CogniteLogger):
                     self.restart()
 
             finally:
-                # Record task end and clear any per-run cancellation token
+                # Record task end
                 self._checkin_worker.report_task_end(name=task.name, timestamp=now())
-                self._running_task_tokens.pop(task.name, None)
 
         task.target = run_task
         self._tasks.append(task)
@@ -502,7 +501,12 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         """
         child_token = self.cancellation_token.create_child_token()
         self._running_task_tokens[task.name] = child_token
-        task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
+        try:
+            task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
+        finally:
+            # Only remove our token; a concurrent run may have already replaced it
+            if self._running_task_tokens.get(task.name) is child_token:
+                self._running_task_tokens.pop(task.name, None)
 
     def start(self) -> None:
         """
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -516,6 +516,9 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             self._running_task_tokens[task.name] = child_token
         try:
             task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
+        except Exception:
+            self._logger.exception(f"Task '{task.name}' raised an unhandled exception")
+            raise
         finally:
             with self._running_task_tokens_lock:
                 if self._running_task_tokens.get(task.name) is child_token:
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -463,6 +463,9 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         Args:
             task: The task to add. It should be an instance of ``StartupTask``, ``ContinuousTask``, or ``ScheduledTask``
         """
+        if any(t.name == task.name for t in self._tasks):
+            raise ValueError(f"Task '{task.name}' is already registered.")
+
         # Store this for later, since we'll override it with the wrapped version
         target = task.target
 
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -85,6 +85,8 @@ from cognite.extractorutils.unstable.configuration.models import (
     create_state_store,
 )
 from cognite.extractorutils.unstable.core._dto import (
+    ActionType,
+    AvailableActionWrite,
     CogniteModel,
     ExtractorInfo,
     StartupRequest,
@@ -94,6 +96,7 @@ from cognite.extractorutils.unstable.core._dto import (
     Task as DtoTask,
 )
 from cognite.extractorutils.unstable.core._messaging import RuntimeMessage
+from cognite.extractorutils.unstable.core.actions import CustomAction
 from cognite.extractorutils.unstable.core.checkin_worker import CheckinWorker
 from cognite.extractorutils.unstable.core.errors import Error, ErrorLevel
 from cognite.extractorutils.unstable.core.logger import CogniteLogger, RobustFileHandler
@@ -106,6 +109,7 @@ __all__ = [
     "CogniteModel",
     "ConfigRevision",
     "ConfigType",
+    "CustomAction",
     "Extractor",
 ]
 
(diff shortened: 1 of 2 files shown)


cognite/extractorutils/unstable/core/base.py
```diff
@@ -184,6 +188,8 @@ def __init__(self, config: FullConfig[ConfigType], checkin_worker: CheckinWorker
         self._scheduler = TaskScheduler(self.cancellation_token.create_child_token())
 
         self._tasks: list[Task] = []
+        self._custom_actions: list[CustomAction] = []
+        self._running_task_tokens: dict[str, CancellationToken] = {}
```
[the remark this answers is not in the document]
vikramlc-cognite: Cleanup was moved from run_task's finally (where child_token is unknown) into _run_task_with_token's finally (where it is in scope), using an identity check:
```
finally:
      if self._running_task_tokens.get(task.name) is child_token:
          self._running_task_tokens.pop(task.name, None)
```

cognite/extractorutils/unstable/core/base.py
```diff
@@ -434,8 +477,9 @@ def run_task(task_context: TaskContext) -> None:
                     self.restart()
 
             finally:
-                # Record task end
+                # Record task end and clear any per-run cancellation token
                 self._checkin_worker.report_task_end(name=task.name, timestamp=now())
+                self._running_task_tokens.pop(task.name, None)
```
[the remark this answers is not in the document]
vikramlc-cognite: [Above](https://github.com/cognitedata/python-extractor-utils/pull/536#discussion_r3332016349)

cognite/extractorutils/unstable/core/base.py
```diff
@@ -445,15 +489,21 @@ def run_task(task_context: TaskContext) -> None:
...
+    def _run_task_with_token(self, task: ScheduledTask) -> None:
+        """
+        Create a fresh child cancellation token for a scheduled task run.
+
+        Stores the token in ``_running_task_tokens`` for external cancellation (e.g. a
+        stop_task action), then executes the task. Called by both the scheduler and any
+        future start_task action handler.
+        """
+        child_token = self.cancellation_token.create_child_token()
+        self._running_task_tokens[task.name] = child_token
```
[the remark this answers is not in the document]
vikramlc-cognite: Fixed [above](https://github.com/cognitedata/python-extractor-utils/pull/536#discussion_r3332016349)

tests/test_unstable/test_action_registration.py
```diff
@@ -0,0 +1,175 @@
...
+    assert "the-task" not in extractor._running_task_tokens
+
+
+def test_scheduled_task_token_is_child_of_extractor_cancellation_token() -> None:
+    extractor = _make_extractor()
+    captured: list = []
+    done = Event()
+
+    def target(_: TaskContext) -> None:
+        captured.append(extractor._running_task_tokens.get("the-task"))
```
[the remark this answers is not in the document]
vikramlc-cognite: The data structure didn't change so no test changes are needed.

cognite/extractorutils/unstable/core/base.py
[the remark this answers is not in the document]
vikramlc-cognite: Added `self._running_task_tokens_lock = RLock()` alongside the `_running_task_tokens` dict

cognite/extractorutils/unstable/core/base.py
```diff
@@ -335,6 +342,25 @@ def __init_tasks__(self) -> None:
...
+        pass
+
+    def add_action(self, action: CustomAction) -> None:
+        """
+        Register a custom action that Odin can invoke on this extractor.
+
+        Args:
+            action: The custom action to register.
+        """
+        self._custom_actions.append(action)
```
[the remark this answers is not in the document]
vikramlc-cognite: Now raises ValueError if a custom action name is already registered or conflicts with an auto-generated "Start {task}" / "Stop {task}" name.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -445,15 +488,26 @@ def run_task(task_context: TaskContext) -> None:
...
+        future start_task action handler.
+        """
+        child_token = self.cancellation_token.create_child_token()
+        self._running_task_tokens[task.name] = child_token
+        try:
+            task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
+        finally:
+            # Only remove our token; a concurrent run may have already replaced it
+            if self._running_task_tokens.get(task.name) is child_token:
+                self._running_task_tokens.pop(task.name, None)
```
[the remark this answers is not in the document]
vikramlc-cognite: Both the token assignment and the identity-check-and-pop are now wrapped in `self._running_task_tokens_lock`, eliminating the TOCTOU race where Thread A's pop could remove Thread B's newly-stored token.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -445,15 +500,27 @@ def run_task(task_context: TaskContext) -> None:
...
+        """
+        child_token = self.cancellation_token.create_child_token()
+        with self._running_task_tokens_lock:
+            self._running_task_tokens[task.name] = child_token
+        try:
+            task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
+        finally:
+            with self._running_task_tokens_lock:
+                if self._running_task_tokens.get(task.name) is child_token:
+                    self._running_task_tokens.pop(task.name, None)
```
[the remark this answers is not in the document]
vikramlc-cognite: These are already taken care of:

> - Orphaned tokens if task runs concurrently
> **TaskScheduler** already prevents overlapping runs (**_scheduler.py:57-59** skips with a warning if the job is already running). The only other concurrent trigger path is a **start_task** action handler, which is future work - we shouldn't design token tracking around it now.
> 
> - Check **is_cancelled** before starting a task
> Redundant. A child of an already cancelled token inherits that state immediately, and the scheduler's own child token stops new jobs from firing during shutdown.
> 
> - **cancel_task(task_name: str) -> bool** public method
> No caller exists yet. The right API shape (return type, ownership, semantics) should be decided when the **stop_task** handler is implemented, not speculatively now.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -445,15 +500,27 @@ def run_task(task_context: TaskContext) -> None:
...
+
+        Stores the token in ``_running_task_tokens`` for external cancellation (e.g. a
+        stop_task action), then executes the task. Called by both the scheduler and any
+        future start_task action handler.
+        """
+        child_token = self.cancellation_token.create_child_token()
+        with self._running_task_tokens_lock:
+            self._running_task_tokens[task.name] = child_token
+        try:
+            task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
```
Toshad, line 529: This will just swallow all exceptions. We should at least log them.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -184,6 +188,9 @@ def __init__(self, config: FullConfig[ConfigType], checkin_worker: CheckinWorker
         self._scheduler = TaskScheduler(self.cancellation_token.create_child_token())
 
         self._tasks: list[Task] = []
+        self._custom_actions: list[CustomAction] = []
```
Toshad, line 191: Why do we have a list for 'custom' actions. Why not a single list for all actions?

vikramlc-cognite: We just have custom action list. **Task** and **CustomAction** are fundamentally different types - **Task** has a schedule, cancellation token lifecycle, and task context; **CustomAction** is a plain callback with a different signature. The **_tasks** list predates this PR entirely.

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: try/finally does not swallow exceptions - exceptions propagate normally after the finally block runs. If **task.target()** raises, the exception travels through **_run_task_with_token**'s finally, then through the scheduler's own wrap() function in **_scheduler.py** (which also has only try/finally, no except), and ends up unhandled in the daemon thread. Python's default threading.excepthook prints it to stderr - not to your logger.
Added a logger for exception like below:
```
except Exception:
    self._logger.exception(f"Task '{task.name}' raised an unhandled exception")
    raise
```

Yaseen-A-Khan: I think the except block in `_run_task_with_token` is unreachable for task exception
is it fine to retain it?

vikramlc-cognite: Thanks for pointing out @Yaseen-A-Khan . It is indeed not reachable. The only way the except in **_run_task_with_token** could fire is if **_checkin_worker.report_task_start/end** raises - infrastructure exceptions, not task exceptions. Reverted the except block addition.

Yaseen-A-Khan (dismissed):
LGTM

cognite/extractorutils/unstable/core/base.py
```diff
@@ -335,6 +343,36 @@ def __init_tasks__(self) -> None:
...
+        reserved_names = {
+            name for t in self._tasks if isinstance(t, ScheduledTask) for name in (f"Start {t.name}", f"Stop {t.name}")
+        }
+        if action.name in reserved_names:
+            raise ValueError(
+                f"Custom action name '{action.name}' conflicts with an auto-generated scheduled task action."
+            )
+
+        self._custom_actions.append(action)
+
```
nithinb, line 375: while you check for duplicates for actions, we do not check for duplicates in add_task? 
I know that changes was made outside of this PR but worth checking

tests/test_unstable/test_action_registration.py
```diff
@@ -0,0 +1,221 @@
...
+        task_running.set()
+        token_present.append("the-task" in extractor._running_task_tokens)
+        allow_finish.wait(timeout=5)
+
+    extractor.add_task(ScheduledTask.from_interval(interval="1h", name="the-task", target=target))
+    extractor._scheduler.trigger("the-task")
+    task_running.wait(timeout=5)
+    allow_finish.set()
+
+    assert token_present == [True]
```
nithinb, line 171: it is quite possible that assertion fails because the target did not finish executing. In my opinion, this will introduce some flakiness

nithinb: Maybe something like this? 


```suggestion

def test_token_present_in_running_task_tokens_during_execution() -> None:
    extractor = _make_extractor()
    token_present: list[bool] = []
    task_running = Event()
    appended = Event()
    allow_finish = Event()

    def target(_: TaskContext) -> None:
        task_running.set()
        token_present.append("the-task" in extractor._running_task_tokens)
        appended.set()
        allow_finish.wait(timeout=5)

    extractor.add_task(ScheduledTask.from_interval(interval="1h", name="the-task", target=target))
    extractor._scheduler.trigger("the-task")
    task_running.wait(timeout=5)
    allow_finish.set()
    appended.wait(timeout=5)
    assert token_present == [True]

```

tests/test_unstable/test_action_registration.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def target(_: TaskContext) -> None:
+        # Simulate a subsequent invocation overwriting the entry before this run finishes
+        extractor._running_task_tokens["the-task"] = replacement
+        done.set()
+
+    extractor.add_task(ScheduledTask.from_interval(interval="1h", name="the-task", target=target))
+    extractor._scheduler.trigger("the-task")
+    done.wait(timeout=5)
+    time.sleep(0.05)  # allow _run_task_with_token's finally to run
```
nithinb: I feel like this will introduce flakiness in tests. 50ms to let `finally` run is fragile in my opinion

nithinb (changes requested):
I have a few comments

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Added duplicate check.

tests/test_unstable/test_action_registration.py
vikramlc-cognite: The race was real: task_running.set() fires before token_present.append(), so the test thread could assert before the append happened. Added appended = Event() (set after the append in the task, waited on in the test before asserting).

tests/test_unstable/test_action_registration.py
vikramlc-cognite: - **test_token_removed**: replaced sleep(0.05) with a poll on **_running_task_tokens** - we're checking it gets removed, so polling that exact condition with a 5s timeout is both correct and self-documenting.
- **test_token_cleanup**: replaced sleep(0.05) with a poll on **_scheduler._running** - the scheduler removes the job from that set only after **_run_task_with_token's** finally completes, making it the definitive "thread is done" signal.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -408,6 +463,9 @@ def add_task(self, task: Task) -> None:
         Args:
             task: The task to add. It should be an instance of ``StartupTask``, ``ContinuousTask``, or ``ScheduledTask``
         """
+        if any(t.name == task.name for t in self._tasks):
+            raise ValueError(f"Task '{task.name}' is already registered.")
+
         # Store this for later, since we'll override it with the wrapped version
```
nithinb, line 477: @vikramlc this still doesn't prevent something like this

```
add_action(CustomAction(name="Start foo"))
add_task(ScheduledTask(name="foo"))
```

nithinb (changes requested):
I think the change made for duplicates will not always work. Possible to take a look at that?

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Comment is valid because **add_task** can technically be called at any time, not just inside **__init_tasks__**. Nothing prevents calling it after add_action, and there's no runtime enforcement of the intended ordering.

nithinb (approved):
risk review ok

PR #538, opened 2026-06-04
feat(odin): Execute incoming actions from Odin

vikramlc-cognite:
### PR Short Summary
Wires the Extractor as the action dispatcher for **CheckinWorker** and implements the full action routing pipeline: **_handle_actions** fans out each incoming action to a dedicated daemon thread, _dispatch_single_action routes by matching the action name against registered available actions, and the three handler methods (**_handle_start_task_action**, **_handle_stop_task_action**, **_handle_custom_action**) execute the action and report lifecycle status updates back through the checkin worker.

### Type of change

### What changed
cognite/extractorutils/unstable/core/base.py
- Added imports: Action, ActionStatus, ActionUpdate from _dto; ActionContext from actions
- _handle_actions(actions) — spawns one daemon thread per action named Action-{external_id} so a slow action never blocks the next
- _dispatch_single_action(action) — builds O(1) sets of known start/stop/custom action names at dispatch time and routes accordingly; any unrecognised name immediately reports failed with a descriptive message (no exception raised)
- _handle_start_task_action(action) — checks task existence, holds _running_task_tokens_lock during the already-running guard, then reports running
→ _run_task_with_token → succeeded/failed
- _handle_stop_task_action(action) — snapshots the child token under lock, cancels it, reports canceled; reports failed if the task is not currently running
- _handle_custom_action(action) — builds ActionContext with call_metadata, invokes custom.target(ctx), reports running → succeeded/failed
- start() — calls self._checkin_worker.set_action_dispatcher(self._handle_actions) before the checkin thread is spawned so no pending actions can arrive without a dispatcher

tests/test_unstable/test_action_dispatch.py — new file, 17 tests covering all acceptance criteria

### Why it changed
Odin dispatches pending actions in the checkin response. Without this wiring, actions are received by CheckinWorker but never executed. This completes the round-trip: the extractor announces capabilities at startup (Task 5) and can now act on them.

- Dispatch key design: type is inferred by matching action_name against the three known sets (start/stop/custom) rather than a type field on Action (which the DTO does not carry). Custom actions named "Start X" where X is not a registered ScheduledTask are legal and route correctly to the custom handler.
- **_handle_start_task_action** concurrency: the already-running check holds **_running_task_tokens_lock** for the read. There is a narrow TOCTOU window between the lock release and **_run_task_with_token** acquiring its own lock; this matches the existing scheduler behaviour and is intentional - **_run_task_with_token's** identity-check finally already handles concurrent overwrites correctly.
- start() ordering: dispatcher is registered before the checkin thread starts — verified by **test_dispatcher_registered_before_checkin_thread_starts**.

### Test evidence
tests/test_unstable/test_action_dispatch.py    17 passed
tests/test_unstable/test_action_registration.py  15 passed
tests/test_unstable/test_actions.py             10 passed
ruff check — All checks passed

### Risks and unknowns
- The "Start {t.name}" / "Stop {t.name}" name convention is shared between _get_startup_request (registration) and _dispatch_single_action (routing). A single source-of-truth constant would eliminate future drift if the convention changes.
- _run_task_with_token's inner run_task wrapper catches all Exceptions internally, so _handle_start_task_action will almost always report succeeded even when the task fails internally (the failure is reported via the error pipeline). This is consistent with the framework's design but means the action status alone does not indicate task success.

### Rollout and rollback
- Purely additive. Extractors that do not register scheduled tasks or custom actions receive no _handle_actions calls and behave identically to before. The dispatcher is only invoked when the server returns pending_actions.

### Checklist
- [x] Self-reviewed the diff
- [x] Tests added or updated
- [x] Docs updated (N/A — internal framework)
- [x] Breaking changes called out above and communicated to affected teams (none)

### DIFF
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -85,7 +85,10 @@ from cognite.extractorutils.unstable.configuration.models import (
     create_state_store,
 )
 from cognite.extractorutils.unstable.core._dto import (
+    Action,
+    ActionStatus,
     ActionType,
+    ActionUpdate,
     AvailableActionWrite,
     CogniteModel,
     ExtractorInfo,
@@ -96,7 +99,7 @@ from cognite.extractorutils.unstable.core._dto import (
     Task as DtoTask,
 )
 from cognite.extractorutils.unstable.core._messaging import RuntimeMessage
-from cognite.extractorutils.unstable.core.actions import CustomAction
+from cognite.extractorutils.unstable.core.actions import ActionContext, CustomAction
 from cognite.extractorutils.unstable.core.checkin_worker import CheckinWorker
 from cognite.extractorutils.unstable.core.errors import Error, ErrorLevel
 from cognite.extractorutils.unstable.core.logger import CogniteLogger, RobustFileHandler
@@ -524,6 +527,136 @@ class Extractor(Generic[ConfigType], CogniteLogger):
                 if self._running_task_tokens.get(task.name) is child_token:
                     self._running_task_tokens.pop(task.name, None)
 
+    def _handle_actions(self, actions: list[Action]) -> None:
+        for action in actions:
+            Thread(
+                target=self._dispatch_single_action,
+                args=(action,),
+                name=f"Action-{action.external_id}",
+                daemon=True,
+            ).start()
+
+    def _dispatch_single_action(self, action: Action) -> None:
+        scheduled_start_names = {f"Start {t.name}" for t in self._tasks if isinstance(t, ScheduledTask)}
+        scheduled_stop_names = {f"Stop {t.name}" for t in self._tasks if isinstance(t, ScheduledTask)}
+        custom_names = {a.name for a in self._custom_actions}
+
+        if action.action_name in scheduled_start_names:
+            self._handle_start_task_action(action)
+        elif action.action_name in scheduled_stop_names:
+            self._handle_stop_task_action(action)
+        elif action.action_name in custom_names:
+            self._handle_custom_action(action)
+        else:
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(
+                    external_id=action.external_id,
+                    status=ActionStatus.failed,
+                    result_message=f"No action named '{action.action_name}' registered",
+                )
+            )
+
+    def _handle_start_task_action(self, action: Action) -> None:
+        task_name = action.action_name[len("Start ") :]
+        task = next(
+            (t for t in self._tasks if t.name == task_name and isinstance(t, ScheduledTask)),
+            None,
+        )
+        if task is None:
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(
+                    external_id=action.external_id,
+                    status=ActionStatus.failed,
+                    result_message=f"No scheduled task named '{task_name}' found",
+                )
+            )
+            return
+
+        with self._running_task_tokens_lock:
+            if task_name in self._running_task_tokens:
+                self._checkin_worker.queue_action_update(
+                    ActionUpdate(
+                        external_id=action.external_id,
+                        status=ActionStatus.failed,
+                        result_message=f"Task '{task_name}' is already running",
+                    )
+                )
+                return
+
+        self._checkin_worker.queue_action_update(
+            ActionUpdate(external_id=action.external_id, status=ActionStatus.running)
+        )
+
+        try:
+            self._run_task_with_token(task)
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(external_id=action.external_id, status=ActionStatus.succeeded)
+            )
+        except Exception as e:
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(
+                    external_id=action.external_id,
+                    status=ActionStatus.failed,
+                    result_message=str(e),
+                )
+            )
+
+    def _handle_stop_task_action(self, action: Action) -> None:
+        task_name = action.action_name[len("Stop ") :]
+        with self._running_task_tokens_lock:
+            token = self._running_task_tokens.get(task_name)
+        if token is None:
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(
+                    external_id=action.external_id,
+                    status=ActionStatus.failed,
+                    result_message=f"Task '{task_name}' is not currently running",
+                )
+            )
+            return
+
+        token.cancel()
+        self._checkin_worker.queue_action_update(
+            ActionUpdate(external_id=action.external_id, status=ActionStatus.canceled)
+        )
+
+    def _handle_custom_action(self, action: Action) -> None:
+        custom = next((a for a in self._custom_actions if a.name == action.action_name), None)
+        if custom is None:
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(
+                    external_id=action.external_id,
+                    status=ActionStatus.failed,
+                    result_message=f"No custom action named '{action.action_name}' registered",
+                )
+            )
+            return
+
+        self._checkin_worker.queue_action_update(
+            ActionUpdate(external_id=action.external_id, status=ActionStatus.running)
+        )
+
+        ctx = ActionContext(
+            action=custom,
+            extractor=self,
+            external_id=action.external_id,
+            call_metadata=action.call_metadata,
+        )
+
+        try:
+            custom.target(ctx)
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(external_id=action.external_id, status=ActionStatus.succeeded)
+            )
+        except Exception as e:
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(
+                    external_id=action.external_id,
+                    status=ActionStatus.failed,
+                    result_message=str(e),
+                )
+            )
+
     def start(self) -> None:
         """
         Start the extractor.
@@ -536,6 +669,8 @@ class Extractor(Generic[ConfigType], CogniteLogger):
 
         self._load_state_store()
         self.state_store.start()
+
+        self._checkin_worker.set_action_dispatcher(self._handle_actions)
 
         Thread(target=self._run_checkin, name="ExtractorCheckin", daemon=True).start()
         if self.metrics_push_manager:
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -516,9 +516,6 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             self._running_task_tokens[task.name] = child_token
         try:
             task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
-        except Exception:
-            self._logger.exception(f"Task '{task.name}' raised an unhandled exception")
-            raise
         finally:
             with self._running_task_tokens_lock:
                 if self._running_task_tokens.get(task.name) is child_token:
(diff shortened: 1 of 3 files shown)


cognite/extractorutils/unstable/core/base.py
```diff
@@ -408,6 +466,9 @@ def add_task(self, task: Task) -> None:
         Args:
             task: The task to add. It should be an instance of ``StartupTask``, ``ContinuousTask``, or ``ScheduledTask``
         """
+        if any(t.name == task.name for t in self._tasks):
+            raise ValueError(f"Task '{task.name}' is already registered.")
```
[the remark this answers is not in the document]
vikramlc-cognite: Valid suggestion. Add this change.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -445,14 +506,156 @@ def run_task(task_context: TaskContext) -> None:
...
+        """
+        child_token = self.cancellation_token.create_child_token()
+        with self._running_task_tokens_lock:
+            self._running_task_tokens[task.name] = child_token
+        try:
+            task.target(TaskContext(task=task, extractor=self, cancellation_token=child_token))
+        finally:
+            with self._running_task_tokens_lock:
+                if self._running_task_tokens.get(task.name) is child_token:
+                    self._running_task_tokens.pop(task.name, None)
```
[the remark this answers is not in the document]
vikramlc-cognite: Already taken care as part of above comment.

Yaseen-A-Khan (commented):
/gemini review

tests/test_unstable/test_action_dispatch.py
```diff
@@ -0,0 +1,383 @@
...
+
+def test_start_task_action_raises_reports_failed() -> None:
+    """If _run_task_with_token itself raises (e.g. child token creation fails), failed is reported."""
+    extractor = _make_extractor()
+    extractor.add_task(ScheduledTask.from_interval(interval="1h", name="boom", target=lambda _: None))
+
+    original = extractor._run_task_with_token
+
+    def raise_on_call(task: ScheduledTask) -> None:
+        raise RuntimeError("token failure")
```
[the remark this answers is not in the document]
Yaseen-A-Khan: I dont find this method, was this removed?

cognite/extractorutils/unstable/core/base.py
```diff
@@ -445,15 +517,164 @@ def run_task(task_context: TaskContext) -> None:
...
+        future start_task action handler.
+
+        If ``child_token`` is supplied (action dispatch path), the caller has already
+        registered it in ``_running_task_tokens`` atomically; this method skips the
+        registration step and goes straight to execution.
+        """
+        if child_token is None:
+            child_token = self.cancellation_token.create_child_token()
+            with self._running_task_tokens_lock:
+                self._running_task_tokens[task.name] = child_token
```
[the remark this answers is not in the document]
vikramlc-cognite: Nice suggestion. Added this change.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -445,15 +517,164 @@ def run_task(task_context: TaskContext) -> None:
...
+
+        self._checkin_worker.queue_action_update(
+            ActionUpdate(external_id=action.external_id, status=ActionStatus.running)
+        )
+
+        try:
+            self._run_task_with_token(task, child_token)
+            self._checkin_worker.queue_action_update(
+                ActionUpdate(external_id=action.external_id, status=ActionStatus.succeeded)
+            )
```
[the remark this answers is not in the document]
vikramlc-cognite: Nice. Added this change.

tests/test_unstable/test_action_dispatch.py
```diff
@@ -0,0 +1,249 @@
...
+    assert "DoesNotExist" in (updates[0].result_message or "")
+
+
+@pytest.mark.parametrize(
+    "register,action_name,expected_status",
+    [
+        pytest.param(
+            lambda ext: ext.add_task(ScheduledTask.from_interval(interval="1h", name="sync", target=lambda _: None)),
+            "Start sync",
+            ActionStatus.running,
```
Yaseen-A-Khan, line 51: Can we also test for the "Stop sync" routing

cognite/extractorutils/unstable/core/base.py
```diff
@@ -506,24 +520,163 @@ def run_task(task_context: TaskContext) -> None:
...
+                    status=ActionStatus.failed,
+                    result_message=f"Task '{task_name}' is not currently running",
+                )
+            )
+            return
+
+        token.cancel()
+        self._checkin_worker.queue_action_update(
+            ActionUpdate(external_id=action.external_id, status=ActionStatus.canceled)
+        )
```
Yaseen-A-Khan, line 640: the window between the releasing the lock and the cancelling, the task may finish,
then the `token.cancel()` is only executing on the completed run
check the token's state after cancellation or whether it was still the live token

cognite/extractorutils/unstable/core/base.py
```diff
@@ -506,24 +520,163 @@ def run_task(task_context: TaskContext) -> None:
...
+                return
+            self._running_task_tokens[task_name] = child_token
+
+        self._checkin_worker.queue_action_update(
+            ActionUpdate(external_id=action.external_id, status=ActionStatus.running)
+        )
+
+        try:
+            self._run_task_with_token(task, child_token)
+            status = ActionStatus.canceled if child_token.is_cancelled else ActionStatus.succeeded
```
Toshad, line 610: Are we sure that token being canceled doesn't throw an exception(it throws exception in asyncio)?
And could add a unit test for cancelled actions as well?

Toshad: Nvm, test exists

Toshad (dismissed):

cognite/extractorutils/unstable/core/base.py
```diff
@@ -506,24 +520,163 @@ def run_task(task_context: TaskContext) -> None:
...
+                )
+            )
+
+    def _handle_start_task_action(self, action: Action) -> None:
+        task_name = action.action_name[len("Start ") :]
+        task = next(
+            (t for t in self._tasks if t.name == task_name and isinstance(t, ScheduledTask)),
+            None,
+        )
+        if task is None:
```
Yaseen-A-Khan, line 581: task is None in `_handle_start_task_action` and custom is None in `_handle_custom_action are unreachable`, `_dispatch_single_action` already verified the name exists before routing here. could remove them or add a comment that these are only reachable if the handlers are called directly.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -506,24 +520,163 @@ def run_task(task_context: TaskContext) -> None:
...
+    def _handle_actions(self, actions: list[Action]) -> None:
+        for action in actions:
+            Thread(
+                target=self._dispatch_single_action,
+                args=(action,),
+                name=f"Action-{action.external_id}",
+                daemon=True,
+            ).start()
+
+    def _dispatch_single_action(self, action: Action) -> None:
```
Yaseen-A-Khan, line 555: `_dispatch_single_action` rebuilds three name sets on every call. Since tasks and actions are immutable after start(), 
is there anyway we can use caching here?

Yaseen-A-Khan (changes requested):
I have some comments regarding the changes

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Technically correct that **_dispatch_single_action** won't reach this path - but **_handle_start_task_action** is a private method, not a sealed one. Tests call it directly, and future code could too. The **None** guard is three lines, gives a meaningful error message, and costs nothing at runtime. Removing it gains nothing, keeping it protects against any caller that bypasses **_dispatch_single_action**. Same reasoning applies to the guard in **_handle_custom_action**.

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Actions are dispatched at most once per checkin cycle. Building three sets over a list of tasks would be in microseconds. Caching would be just an overkill here at this point.

tests/test_unstable/test_action_dispatch.py
vikramlc-cognite: Stop routing is already fully covered - **test_stop_task_action_not_running_reports_failed** and **test_stop_task_action_cancels_child_token_and_reports_canceled** test both failure and success paths end-to-end.

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Valid. Fixed.

andersfylling:
once team has approved, ping me on slack and I'll do the risk review ASAP =)
Removing waiting on risk review, since a team approval is required.

Yaseen-A-Khan (approved):

Yaseen-A-Khan (dismissed):

cognite/extractorutils/unstable/core/base.py
```diff
@@ -514,24 +517,165 @@ def run_task(task_context: TaskContext) -> None:
...
 
         Stores the token in ``_running_task_tokens`` for external cancellation (e.g. a
         stop_task action), then executes the task. Called by both the scheduler and any
         future start_task action handler.
+
+        If ``child_token`` is supplied (action dispatch path), the caller has already
+        registered it in ``_running_task_tokens`` atomically; this method skips the
+        registration step and goes straight to execution.
         """
-        child_token = self.cancellation_token.create_child_token()
```
andersfylling, line 525: Are task names guaranteed unique?

cognite/extractorutils/unstable/core/base.py
```diff
@@ -514,24 +517,165 @@ def run_task(task_context: TaskContext) -> None:
...
-        with self._running_task_tokens_lock:
-            self._running_task_tokens[task.name] = child_token
+        if child_token is None:
+            with self._running_task_tokens_lock:
+                if task.name in self._running_task_tokens:
+                    self._logger.warning("Task '%s' is already running, skipping scheduled run.", task.name)
+                    return
+                child_token = self.cancellation_token.create_child_token()
+                self._running_task_tokens[task.name] = child_token
```
andersfylling, line 538: my concern here is that previously if name were not unique it would override the cancellation token, and older tasks would gain the new cancellation token behavior. Now it's the other way around; new tasks will inherit existing tokens.

I think you should considering using a list and be certain you clean it up after a task has finished. Or ensure you have a unique identifier, if name isn't already.

andersfylling (changes requested):

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Yes they are. **add_task** at line 469 enforces uniqueness:
```
if any(t.name == task.name for t in self._tasks):
    raise ValueError(f"Task '{task.name}' is already registered.")
```
It raises on registration, so **_running_task_tokens** can never have two different **ScheduledTask** objects contending for the same name key. Using task.name as the dict key is safe by construction.

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Task names are already unique - **add_task** raises if you try to register a duplicate. A scheduled task is also only meant to have one run at a time, so skipping if already running is intentional. 
Cleanup is handled in the finally block with an identity check (**is child_token**), so even in edge cases a stale entry won't be removed by the wrong run.

andersfylling (dismissed):
🦄

andersfylling (approved):
🦄

Yaseen-A-Khan (approved):

PR #541, opened 2026-06-19
feat(odin): Register fetch logs action and validate input dates

vikramlc-cognite:
### Summary
Implements the first step of the bulk log upload feature for Odin: registers a built-in **fetch_logs** action on every extractor instance and validates the incoming date-range parameters before any file I/O is attempted.

---
### Type of change

---
### What changed
- **ActionError** exception class (**actions.py**): New exception for structured action failures. Carries **error_type** (machine-readable category) and optional details, both surfaced as **result_metadata** in the Odin action update. **_handle_custom_action** now catches ActionError before the generic Exception clause so structured metadata is preserved on the update.
- **_log_upload_action.py** (new private module): Contains **fetch_logs_action** - validates start_date / end_date from `ctx.call_metadata` (ISO 8601, inclusive, max 7 days). Raises ActionError with error_type of **missing_parameter**, invalid_parameter, or invalid_date_range on any validation failure.
- `Extractor._register_builtin_actions()` (base.py): New private method called unconditionally in `__init__` (between `__init_tasks__` and `__init_actions__`). Registers **fetch_logs** on every extractor without requiring subclass opt-in. add_action now raises **ValueError** on duplicate name to prevent subclasses from shadowing built-ins.

---
### Why it changed
Operators need to trigger log retrieval from the Odin UI without any per-extractor code changes. Registering as a built-in ensures the action is available on all extractors uniformly.

---
### Related docs / discussion: 
[Design Doc - Bulk Log Upload Action](https://cognitedata.atlassian.net/wiki/spaces/~7120207333078ba8a4415f98a7c58c2f4877a8/pages/6099992944/Design+Doc+Bulk+Log+Upload+Action+all+extractors+support)

---
- **_register_builtin_actions** is called in `__init__` before `__init_actions__`. Confirm this ordering is safe for all existing extractor subclasses (subclasses that call add_action in `__init_actions__` will see **fetch_logs** already present).
- The except **ActionError** branch in **_handle_custom_action** must sit before except Exception - any reordering silently swallows structured metadata.
- add_action duplicate-name guard: raises ValueError at registration time, not at dispatch time. Intentional - fail fast rather than silently overwrite.

---
### Test evidence
- `tests/test_unstable/test_log_upload_action.py` (new, 11 test cases): registration, all missing-param combinations, invalid ISO formats, invalid date ranges, boundary at exactly 7 days, ActionError.details wiring.
- `tests/test_unstable/test_action_dispatch.py`: added `test_action_error_sets_result_metadata_and_keeps_failed_status` covering the except ActionError branch in dispatch.
- `tests/test_unstable/test_action_registration.py`: updated 6 existing tests to be semantic (name/type lookups) rather than count-based, since the built-in adds one action to every extractor.
- All 43 tests pass; ruff clean.

---
### Risks and unknowns
- File I/O and CDF Files upload are not yet implemented - **fetch_logs_action** returns successfully after validation. A valid date range call currently does nothing beyond parameter checking. Subsequent PRs complete the implementation ([Tasks 2–5](https://cognitedata.atlassian.net/browse/EDGE-584)).
- Log file path detection (Task 2) depends on the extractor's configured file handler - the approach for extractors with no file handler is not yet finalised.

---
### Rollout and rollback
- No config changes or API changes required. Rolling back means reverting this commit; no data is written to CDF at this stage. Once [Tasks 2–5](https://cognitedata.atlassian.net/browse/EDGE-584) land, rollback would require ensuring no in-flight uploads are orphaned.

---
### Checklist
- [x] Self-reviewed the diff
- [x] Tests added or updated
- [x] Docs updated - N/A (design doc is pre-existing)
- [x] Breaking changes called out - none

### DIFF
--- a/cognite/extractorutils/unstable/core/_log_upload_action.py
+++ b/cognite/extractorutils/unstable/core/_log_upload_action.py
@@ -1,8 +1,11 @@
 """Built-in ``fetch_logs`` action: streams rotated log files to CDF Files."""
 
+import logging
 from datetime import date
 
 from cognite.extractorutils.unstable.core.actions import ActionContext, ActionError
+
+_logger = logging.getLogger(__name__)
 
 MAX_DATE_RANGE_DAYS = 7
 """Maximum number of calendar days a single ``fetch_logs`` invocation may cover."""
@@ -50,3 +53,5 @@ def fetch_logs_action(ctx: ActionContext) -> None:
             "use multiple invocations for longer spans",
             error_type="invalid_date_range",
         )
+
+    _logger.info("fetch_logs: uploading logs for %s to %s (%d day(s))", start_date, end_date, num_days)
--- a/cognite/extractorutils/unstable/core/_log_upload_action.py
+++ b/cognite/extractorutils/unstable/core/_log_upload_action.py
@@ -1,0 +1,54 @@
+"""Built-in ``fetch_logs`` action: streams rotated log files to CDF Files."""
+
+from datetime import date
+
+from cognite.extractorutils.unstable.core.actions import ActionContext, ActionError
+
+MAX_DATE_RANGE_DAYS = 7
+"""Maximum number of calendar days a single ``fetch_logs`` invocation may cover."""
+
+_FETCH_LOGS_DESCRIPTION = (
+    "Upload rotated log files to CDF Files for a given date range. "
+    f"At most {MAX_DATE_RANGE_DAYS} days per invocation."
+)
+
+
+def _parse_date(raw: str, field: str) -> date:
+    try:
+        return date.fromisoformat(raw)
+    except ValueError:
+        raise ActionError(
+            f"Invalid {field} '{raw}': expected ISO 8601 date (YYYY-MM-DD)",
+            error_type="invalid_parameter",
+        ) from None
+
+
+def fetch_logs_action(ctx: ActionContext) -> None:
+    """Validate parameters and upload rotated log files for the requested date range to CDF Files."""
+    params = ctx.call_metadata or {}
+
+    start_date_raw = params.get("start_date")
+    end_date_raw = params.get("end_date")
+
+    if start_date_raw is None:
+        raise ActionError("Missing required parameter: start_date", error_type="missing_parameter")
+    if end_date_raw is None:
+        raise ActionError("Missing required parameter: end_date", error_type="missing_parameter")
+
+    start_date = _parse_date(start_date_raw, "start_date")
+    end_date = _parse_date(end_date_raw, "end_date")
+
+    if end_date < start_date:
+        raise ActionError(
+            f"end_date ({end_date}) must be on or after start_date ({start_date})",
+            error_type="invalid_date_range",
+        )
+
+    num_days = (end_date - start_date).days + 1
+    if num_days > MAX_DATE_RANGE_DAYS:
+        raise ActionError(
+            f"Date range of {num_days} days exceeds the maximum of {MAX_DATE_RANGE_DAYS} days; "
+            "use multiple invocations for longer spans",
+            error_type="invalid_date_range",
+        )
+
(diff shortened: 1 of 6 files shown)


cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def _parse_date(raw: str, field: str) -> date:
+    try:
+        return date.fromisoformat(raw)
+    except ValueError:
+        raise ActionError(
+            f"Invalid {field} '{raw}': expected ISO 8601 date (YYYY-MM-DD)",
+            error_type="invalid_parameter",
+        ) from None
```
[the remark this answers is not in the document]
vikramlc-cognite: [Fixed](https://github.com/cognitedata/python-extractor-utils/pull/541/commits/a480f65139b95a320aa2fe4f5a46e0d360cee4d0).

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -0,0 +1,52 @@
...
+            error_type="invalid_date_range",
+        )
+
+    num_days = (end_date - start_date).days + 1
+    if num_days > MAX_DATE_RANGE_DAYS:
+        raise ActionError(
+            f"Date range of {num_days} days exceeds the maximum of {MAX_DATE_RANGE_DAYS} days; "
+            "use multiple invocations for longer spans",
+            error_type="invalid_date_range",
+        )
```
nithinb, line 55: at the end, there should be some log to indicate that it succeeded or just do an ActionError("Not yet implemented") or something like that?

cognite/extractorutils/unstable/core/base.py
```diff
@@ -346,6 +348,16 @@ def __init_tasks__(self) -> None:
...
+        """Register framework-level actions available on every extractor."""
+        self.add_action(
+            CustomAction(
+                name="fetch_logs",
+                target=fetch_logs_action,
+                description=_FETCH_LOGS_DESCRIPTION,
+            )
+        )
+
     def __init_actions__(self) -> None:
```
nithinb, line 361: This is a bit conflicting, its a built-in action being regarded as a custom action, which is a bit weird

tests/test_unstable/test_action_registration.py
```diff
@@ -79,19 +83,22 @@ def test_scheduled_and_custom_actions_combined_ordering() -> None:
     extractor.add_action(CustomAction(name="flush", target=lambda _: None))
     req = _startup_request(extractor)
     assert req.available_actions is not None
-    assert len(req.available_actions) == 3
     names = [a.name for a in req.available_actions]
-    assert names == ["Start sync", "Stop sync", "flush"]
+    # Built-in actions precede scheduled-task start/stop, which precede user custom actions
```
nithinb: this doesn't validate the order of custom actions?

cognite/extractorutils/unstable/core/base.py
nithinb: Maybe we can also have CogniteActions?

nithinb (changes requested):

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Got it. Added a logger at the end.

tests/test_unstable/test_action_registration.py
vikramlc-cognite: The code puts scheduled task start/stop first, then _custom_actions in registration order (built-ins first, user actions after). But agree an assert can be added here. Added.

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: **fetch_logs** uses the identical dispatch mechanism as user-registered actions (callable + ActionContext). The only difference is registration time (automatic in **__init__** vs explicit in **__init_actions__**). Introducing a parallel type (**CogniteActions**) hierarchy for the same runtime behaviour adds complexity without changing anything observable.

nithinb (approved):

erlendvollset (commented):
🦄

PR #542, opened 2026-06-19
feat(odin): Build candidate log file list in Extractor

vikramlc-cognite:
### Summary
Builds the candidate log file list for the fetch_logs built-in action. After validating the date range, the action now resolves the extractor's log file path from config, enumerates the [start_date, end_date] date range, and partitions files into candidates (exist, non-empty) and skipped (missing or empty). This is the second step of the bulk log upload feature - it sets up the file set that Tasks 3–4 will stream to CDF Files.

---
### Type of change

---
### What changed
- `_log_upload_action.py`:
  - **LogFileCandidate** frozen dataclass - carries date, path, is_current. is_current=True marks the live file.log (unrotated), which Task 3 (**BoundedReader**) uses to know a size snapshot is required before reading.
  - `_resolve_log_file_path(config: ExtractorConfig) -> Path | None` - returns the base log path from the first LogFileHandlerConfig in config.log_handlers, or None if no file handler is configured.
  - `_build_candidate_files(base_path, start_date, end_date, today)` - enumerates the date range. Historical dates resolve to <base>.YYYY-MM-DD, today resolves to the live base_path. Uses try/except FileNotFoundError on stat() (not exists() + stat()) to eliminate the TOCTOU race during log rotation. Files that don't exist or are 0 bytes go to skipped.
  - **fetch_logs_action** extended: after date validation, resolves the log path (raises `ActionError(error_type="no_file_handler_configured")` if absent), then calls _build_candidate_files to produce candidates and skipped lists for Tasks 3–5.

---
### Why it changed
Part of the bulk log upload design (**refer section: Related docs**). Before uploading anything, the action needs to know which files exist and are non-empty for the requested date range. Separating file discovery from upload also makes each stage independently testable.

---
### Related docs / discussion: 
[Design Doc - Bulk Log Upload Action](https://cognitedata.atlassian.net/wiki/spaces/~7120207333078ba8a4415f98a7c58c2f4877a8/pages/6099992944/Design+Doc+Bulk+Log+Upload+Action+all+extractors+support)

---
- **_build_candidate_files** uses try/except FileNotFoundError around path.stat() rather than path.exists() + path.stat(). This is intentional - eliminates the TOCTOU window where the file is deleted between the two calls (which can happen during log rotation rename).
- The rotation race scenario: if the action is dispatched immediately after midnight before **TimedRotatingFileHandler** renames file.log → file.log.YYYY-MM-DD, the rotated file won't exist yet and that date goes to skipped. This is documented in **_build_candidate_files**'s docstring. The operator retries after rotation completes (within seconds).
- **_resolve_log_file_path** returns the first **LogFileHandlerConfig** found. Extractors with multiple file handlers will have the first one used. This is consistent with how base.py sets up log handlers.
- **fetch_logs_action** accesses **ctx._extractor.application_config** - acceptable since both are in the private framework implementation. Resolving at call time (not at registration time) ensures config changes are reflected immediately.

---
### Test evidence
- `tests/test_unstable/test_log_upload_action.py` (17 test functions, 26 parametrized cases total):
  - `_resolve_log_file_path`: console-only → None; single file handler; multiple file handlers → first
  - `_build_candidate_files`: all files present; missing/empty files → skipped; today uses live file; mixed range; boundary (single day, exact max)
  - `fetch_logs_action integration`: no file handler → no_file_handler_configured; file handler + files present → succeeded; file handler + all files missing → succeeded (empty upload set)
- All 55 tests pass; ruff clean.

---
### Risks and unknowns
- **_candidates and _skipped** are computed but not yet consumed - Tasks 3–5 will use them. The action currently succeeds after building the list regardless of whether any files were found.
- Log file path detection assumes **LogFileHandlerConfig.path** is the actual path on disk. If the extractor is run in a container with volume mounts that differ from the configured path, files won't be found and all dates will be skipped.
- The rotation race window (sub-second) is handled gracefully by skipping the missing date, but there is no automatic retry - the operator must re-invoke the action.

---
### Rollout and rollback
No config changes or API changes required. **fetch_logs_action** now raises **no_file_handler_configured** for extractors without a file log handler (previously it would have silently succeeded with no uploads - which is the same observable behaviour since uploads aren't implemented yet).
Rolling back means reverting this commit, no CDF state is written at this stage.

---
### Checklist
- [x] Self-reviewed the diff
- [x] Tests added or updated
- [x] Docs updated - N/A (design doc is pre-existing)
- [x] Breaking changes called out - none

### DIFF
--- a/cognite/extractorutils/unstable/core/_log_upload_action.py
+++ b/cognite/extractorutils/unstable/core/_log_upload_action.py
@@ -1,8 +1,11 @@
 """Built-in ``fetch_logs`` action: streams rotated log files to CDF Files."""
 
+import logging
 from datetime import date
 
 from cognite.extractorutils.unstable.core.actions import ActionContext, ActionError
+
+_logger = logging.getLogger(__name__)
 
 MAX_DATE_RANGE_DAYS = 7
 """Maximum number of calendar days a single ``fetch_logs`` invocation may cover."""
@@ -50,3 +53,5 @@ def fetch_logs_action(ctx: ActionContext) -> None:
             "use multiple invocations for longer spans",
             error_type="invalid_date_range",
         )
+
+    _logger.info("fetch_logs: uploading logs for %s to %s (%d day(s))", start_date, end_date, num_days)
--- a/cognite/extractorutils/unstable/core/_log_upload_action.py
+++ b/cognite/extractorutils/unstable/core/_log_upload_action.py
@@ -137,7 +137,7 @@ def fetch_logs_action(ctx: ActionContext) -> None:
             error_type="invalid_date_range",
         )
 
-    log_file_path = _resolve_log_file_path(ctx._extractor.application_config)
+    log_file_path = _resolve_log_file_path(ctx.application_config)
     if log_file_path is None:
         raise ActionError(
             "No file log handler configured; add a 'file' type log handler to enable log uploads",
--- a/cognite/extractorutils/unstable/core/_log_upload_action.py
+++ b/cognite/extractorutils/unstable/core/_log_upload_action.py
@@ -1,7 +1,11 @@
 """Built-in ``fetch_logs`` action: streams rotated log files to CDF Files."""
 
-from datetime import date
+from dataclasses import dataclass
+from datetime import date, timedelta, timezone
+from datetime import datetime as dt
+from pathlib import Path
 
+from cognite.extractorutils.unstable.configuration.models import ExtractorConfig, LogFileHandlerConfig
 from cognite.extractorutils.unstable.core.actions import ActionContext, ActionError
 
 MAX_DATE_RANGE_DAYS = 7
@@ -12,6 +16,19 @@ _FETCH_LOGS_DESCRIPTION = (
 )
 
 
+@dataclass(frozen=True)
+class LogFileCandidate:
+    """A log file resolved for a given date that exists and has content."""
+
+    date: date
+    path: Path
+    is_current: bool  # True when path is the live (unrotated) file.log
+
+
+def _today_utc() -> date:
+    return dt.now(tz=timezone.utc).date()
+
+
 def _parse_date(raw: str, field: str) -> date:
     try:
         return date.fromisoformat(raw)
@@ -20,6 +37,63 @@ def _parse_date(raw: str, field: str) -> date:
             f"Invalid {field} '{raw}': expected ISO 8601 date (YYYY-MM-DD)",
             error_type="invalid_parameter",
         ) from None
+
+
+def _resolve_log_file_path(config: ExtractorConfig) -> Path | None:
+    """Return the base log file path from the first file handler in config, or None."""
+    for handler in config.log_handlers:
+        if isinstance(handler, LogFileHandlerConfig):
+            return handler.path
+    return None
+
+
+def _build_candidate_files(
+    base_path: Path,
+    start_date: date,
+    end_date: date,
+    today: date,
+) -> tuple[list[LogFileCandidate], list[date]]:
+    """
+    Enumerate log files for [start_date, end_date] and partition into candidates vs skipped.
+
+    Rotated files follow the naming convention ``<base_path>.YYYY-MM-DD``.
+    The live file (``base_path``) is used for ``today``; all other dates use the rotated name.
+
+    A file is skipped when it does not exist or is empty. This covers the brief rotation
+    race window: if the action is dispatched right after midnight before ``TimedRotatingFileHandler``
+    has renamed ``file.log`` to ``file.log.YYYY-MM-DD``, the rotated file will not be found and
+    that date will appear in ``skipped``. Retrying after rotation completes (within seconds)
+    will pick it up.
+
+    Returns:
+        (candidates, skipped): candidates are files that exist and have content;
+        skipped contains dates for which no usable file was found.
+    """
+    candidates: list[LogFileCandidate] = []
+    skipped: list[date] = []
+
+    current = start_date
+    while current <= end_date:
+        if current == today:
+            path = base_path
+            is_current = True
+        else:
+            path = base_path.parent / f"{base_path.name}.{current.isoformat()}"
+            is_current = False
+
+        try:
+            size = path.stat().st_size
+        except FileNotFoundError:
+            skipped.append(current)
+        else:
+            if size > 0:
+                candidates.append(LogFileCandidate(date=current, path=path, is_current=is_current))
+            else:
+                skipped.append(current)
+
+        current += timedelta(days=1)
+
+    return candidates, skipped
 
 
 def fetch_logs_action(ctx: ActionContext) -> None:
@@ -50,3 +124,13 @@ def fetch_logs_action(ctx: ActionContext) -> None:
             "use multiple invocations for longer spans",
             error_type="invalid_date_range",
         )
+
+    log_file_path = _resolve_log_file_path(ctx._extractor.application_config)
+    if log_file_path is None:
+        raise ActionError(
+            "No file log handler configured; add a 'file' type log handler to enable log uploads",
+            error_type="no_file_handler_configured",
+        )
+
+    today = _today_utc()
+    _candidates, _skipped = _build_candidate_files(log_file_path, start_date, end_date, today)
--- a/cognite/extractorutils/unstable/core/_log_upload_action.py
+++ b/cognite/extractorutils/unstable/core/_log_upload_action.py
@@ -1,5 +1,6 @@
 """Built-in ``fetch_logs`` action: streams rotated log files to CDF Files."""
 
+import logging
 from dataclasses import dataclass
 from datetime import date, timedelta, timezone
 from datetime import datetime as dt
@@ -7,6 +8,8 @@ from pathlib import Path
 
 from cognite.extractorutils.unstable.configuration.models import ExtractorConfig, LogFileHandlerConfig
 from cognite.extractorutils.unstable.core.actions import ActionContext, ActionError
+
+_logger = logging.getLogger(__name__)
 
 MAX_DATE_RANGE_DAYS = 7
 """Maximum number of calendar days a single ``fetch_logs`` invocation may cover."""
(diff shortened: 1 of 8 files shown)


cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -0,0 +1,136 @@
...
+            path = base_path
+            is_current = True
+        else:
+            path = base_path.parent / f"{base_path.name}.{current.isoformat()}"
+            is_current = False
+
+        try:
+            size = path.stat().st_size
+        except FileNotFoundError:
+            skipped.append(current)
```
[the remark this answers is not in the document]
vikramlc-cognite: [Fixed](https://github.com/cognitedata/python-extractor-utils/pull/542/commits/471e5e15b1c05cb5371def55b7014d1ee584ca58)

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -0,0 +1,136 @@
...
+        raise ActionError("Missing required parameter: end_date", error_type="missing_parameter")
+
+    start_date = _parse_date(start_date_raw, "start_date")
+    end_date = _parse_date(end_date_raw, "end_date")
+
+    if end_date < start_date:
+        raise ActionError(
+            f"end_date ({end_date}) must be on or after start_date ({start_date})",
+            error_type="invalid_date_range",
+        )
```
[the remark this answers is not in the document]
vikramlc-cognite: Valid. Added this change.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -0,0 +1,136 @@
...
+
+    log_file_path = _resolve_log_file_path(ctx._extractor.application_config)
+    if log_file_path is None:
+        raise ActionError(
+            "No file log handler configured; add a 'file' type log handler to enable log uploads",
+            error_type="no_file_handler_configured",
+        )
+
+    today = _today_utc()
+    _candidates, _skipped = _build_candidate_files(log_file_path, start_date, end_date, today)
```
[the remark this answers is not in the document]
vikramlc-cognite: Redundant _today_utc() call: Since today is now computed earlier for the future-date check, the second call at the bottom of fetch_logs_action is removed and the existing variable is reused.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -43,10 +117,27 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
         )
+
+    log_file_path = _resolve_log_file_path(ctx._extractor.application_config)
+    if log_file_path is None:
+        raise ActionError(
+            "No file log handler configured; add a 'file' type log handler to enable log uploads",
+            error_type="no_file_handler_configured",
+        )
+
+    _candidates, _skipped = _build_candidate_files(log_file_path, start_date, end_date, today)
```
nithinb: I presume there is continuation after this? Like I didn't understand why it stops here?

tests/test_unstable/test_log_upload_action.py
```diff
@@ -125,22 +140,6 @@ def test_invalid_date_range_reports_invalid_date_range(
...
-    ids=["single_day", "exact_max_days"],
-)
-def test_valid_date_range_succeeds(start_str: str, end_str: str) -> None:
-    extractor = _make_extractor()
-    updates = _dispatch(extractor, {"start_date": start_str, "end_date": end_str})
-    statuses = [u.status for u in updates]
-    assert ActionStatus.succeeded in statuses
-    assert ActionStatus.failed not in statuses
-
-
```
nithinb, line 143: why was this test removed?

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -22,6 +39,63 @@ def _parse_date(raw: str, field: str) -> date:
...
+        if current == today:
+            path = base_path
+            is_current = True
+        else:
+            path = base_path.parent / f"{base_path.name}.{current.isoformat()}"
+            is_current = False
+
+        try:
+            size = path.stat().st_size
+        except OSError:
```
nithinb: shouldn't we log what the OSError was?

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -22,6 +39,63 @@ def _parse_date(raw: str, field: str) -> date:
...
+        if isinstance(handler, LogFileHandlerConfig):
+            return handler.path
+    return None
+
+
+def _build_candidate_files(
+    base_path: Path,
+    start_date: date,
+    end_date: date,
+    today: date,
```
nithinb, line 57: I didn't really understand the need to take this as a parameter? why not just generate today inside the method?

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -43,10 +117,27 @@ def fetch_logs_action(ctx: ActionContext) -> None:
             error_type="invalid_date_range",
         )
 
+    today = _today_utc()
```
nithinb, line 124: Got the reason for passing today as a parameter.

tests/test_unstable/test_log_upload_action.py
nithinb: exact_max_days no longer has a test

tests/test_unstable/test_log_upload_action.py
```diff
@@ -153,3 +152,120 @@ def raise_with_details(ctx: ActionContext) -> None:
...
+
+def test_today_uses_live_file_not_rotated_name(tmp_path: Path) -> None:
+    base = tmp_path / "extractor.log"
+    base.write_bytes(b"today data")
+    candidates, skipped = _build_candidate_files(base, _PAST_TODAY, _PAST_TODAY, _PAST_TODAY)
+    assert len(candidates) == 1
+    assert candidates[0].path == base
+    assert candidates[0].is_current is True
+    assert skipped == []
+
```
nithinb, line 215: I think it would be useful to have a test for date range of yesterday to today

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -43,10 +117,27 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+
     num_days = (end_date - start_date).days + 1
     if num_days > MAX_DATE_RANGE_DAYS:
         raise ActionError(
             f"Date range of {num_days} days exceeds the maximum of {MAX_DATE_RANGE_DAYS} days; "
             "use multiple invocations for longer spans",
             error_type="invalid_date_range",
         )
+
+    log_file_path = _resolve_log_file_path(ctx._extractor.application_config)
```
nithinb: Nit: One thing to consider here to expose application config using a method or a property since _extractor denotes a private variable

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -12,6 +16,19 @@
 )
 
 
+@dataclass(frozen=True)
+class LogFileCandidate:
+    """A log file resolved for a given date that exists and has content."""
+
+    date: date
```
nithinb: better to avoid this naming. 


```suggestion
    log_data: date
```

nithinb (changes requested):

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Valid. Silent swallowing of OSError makes permission errors invisible in logs. Logged error.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Valid concern. The function ends with candidates/skipped but does nothing with them. Add a log showing results - upload will be in subsequent PR but we should show something. Added a logger.

tests/test_unstable/test_log_upload_action.py
vikramlc-cognite: Valid. This was removed as part of code refactoring. Added it back.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Mentioned [here](https://github.com/cognitedata/python-extractor-utils/pull/542#discussion_r3464676446)

tests/test_unstable/test_log_upload_action.py
vikramlc-cognite: Valid, implemented.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Renamed it to **log_date**

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: **fetch_logs_action** is in **_log_upload_action**.py (framework-internal, _ prefix). Adding **application_config** to **ActionContext**'s public API would expose it to all custom action authors unnecessarily, creating bloat. Internal framework modules accessing each other within the same package should be acceptable. What do you think?

nithinb: > Adding application_config to ActionContext's public API

I did not understand this, how would be expose this to the API? 

A way to solve for this would be either have property defined using the property decorator in python or just define a method in `ActionContext` to get_extractor, and you can do 

```
extractor = ctx.get_extractor()
application_config = extractor.application_config 
```

Do see if there is a better alternative in this approach.

nithinb (changes requested):
Thanks for resolving the rest of the issues. 

Can you take another look at the conversation related to private variables?

cognite/extractorutils/unstable/core/_log_upload_action.py
nithinb: you can every consider this

```
# actions.py — add to ActionContext
@property
def application_config(self) -> "ExtractorConfig":
    return self._extractor.application_config

```

vikramlc-cognite:
@nithinb Agree, my concern was misplaced. I have gone ahead and added the application_config in the action context.

nithinb (approved):
risk review ok

PR #545, opened 2026-06-24
EDGE-614 update renovate.json to use the renovate-config-public repo

Yaseen-A-Khan:
This PR updates the renovate.json which now should point to the repo `renovate-config-public` which defines the renovate settings.


mohammad-arif662 (approved):

thorkildcognite (approved):
🦄 LGTM -- this public repo will not have any dependencies to internal repositores, so makes sense to use the -public variant.

PR #547, opened 2026-06-25
feat(odin): Add support for streaming log files to CDF Files

vikramlc-cognite:
### Summary
Completes the fetch_logs action by wiring the upload loop: candidate log files are streamed to CDF Files via **upload_bytes**, with per-file result tracking and structured success metadata reported back to Odin.

### Type of change

### What changed
- `_log_upload_action.py`: Added **_FileUploadResult dataclass, _file_external_id()**, and **_upload_candidate()**. The **fetch_logs_action** now snapshots the current-day file size before spawning uploads, submits candidates to a **ThreadPoolExecutor**, and calls `ctx.set_result()` with aggregated counts and a per-file JSON array in result_metadata.
- `actions.py`: Added ActionContext.set_result(message, *, metadata) so actions can declare structured success data without changing the ActionTarget signature.
- `base.py`: _handle_custom_action success path now forwards **ctx._result_message and ctx._result_metadata** into the ActionUpdate(succeeded) call.
- `test_log_upload_action.py`: 9 new tests covering **_file_external_id** naming, **_upload_candidate** for rotated files, BoundedReader usage for current-day files, oversized file skipping, CDF upload errors, and end-to-end fetch_logs_action result metadata (partial failures, all-missing, upload failures still report succeeded).
- `test_action_dispatch.py`: 1 new test verifying set_result values propagate into the succeeded ActionUpdate.

### Why it changed
- Implements upload execution and result reporting for the fetch_logs action (EDG-371). Prior commits established parameter validation, candidate file enumeration, and BoundedReader; this commit closes the loop end-to-end.

- Snapshot-before-threads ordering: snapshot_size is captured via stat() before the **ThreadPoolExecutor** is created. This is load-bearing - if the snapshot happened inside the thread, it would race with the logger appending to the file, making BoundedReader's ceiling unreliable.
- Partial failure semantics: A file that fails to upload results in status="failed" in result_metadata["files"], but the action still reports **ActionStatus.succeeded** with failed_files > 0. The caller (Odin operator) sees the breakdown and decides whether to re-trigger.
- ctx._extractor access: fetch_logs_action reads **ctx._extractor.cognite_client and ctx._extractor.connection_config.integration.external_id**. This is internal framework access - `_log_upload_action.py` is _-prefixed and co-located in the same package. Exposing these on the public ActionContext API would unnecessarily widen the surface for user-authored actions.

### Test evidence
`tests/test_unstable/test_log_upload_action.py`  40 passed
`tests/test_unstable/test_action_dispatch.py`    passed
ruff: clean, mypy: clean

### Risks and unknowns
- **DEFAULT_CONCURRENT_UPLOADS = 1** keeps uploads sequential. Safe on constrained networks; parallelism is a one-line change when needed.
- **MAX_FILE_SIZE_BYTES = 4 GiB** - should be validated against the SDK's actual per-request upload limit.

### Rollout and rollback
N/A - triggered only on explicit Odin dispatch. No config changes or migrations.

### Checklist
- [x] Self-reviewed the diff
- [x] Tests added or updated
- [x] Docs updated (N/A - internal framework)
- [x] Breaking changes called out above and communicated to affected teams


cognite/extractorutils/unstable/core/_bounded_reader.py
```diff
@@ -0,0 +1,56 @@
...
+        self._remaining = max_bytes
+
+    def __len__(self) -> int:
+        # super_len() checks __len__ before os.fstat — returning the fixed snapshot
+        # size here ensures Content-Length is declared at the moment of the snapshot,
+        # not at the moment the upload resolves the inode size (which may have grown).
+        return self._size
+
+    def tell(self) -> int:
+        return self._size - self._remaining
```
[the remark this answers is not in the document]
vikramlc-cognite: No longer valid. This file was moved to other PR.

tests/test_unstable/test_bounded_reader.py
```diff
@@ -0,0 +1,103 @@
...
+
+def test_tell_tracks_bytes_consumed(tmp_path: Path) -> None:
+    path = _make_file(tmp_path, b"hello world")
+    with open(path, "rb") as f:
+        reader = BoundedReader(f, 8)
+        assert reader.tell() == 0
+        reader.read(3)
+        assert reader.tell() == 3
+        reader.read(5)
+        assert reader.tell() == 8
```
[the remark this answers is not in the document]
vikramlc-cognite: No longer valid. This file was moved to other PR.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -100,6 +118,68 @@ def _build_candidate_files(
...
+            log_date=candidate.log_date,
+            file_external_id=external_id,
+            status="skipped_too_large",
+            size_bytes=actual_size,
+        )
+
+    try:
+        f = open(candidate.path, "rb")  # noqa: SIM115
+        reader: BinaryIO = BoundedReader(f, snapshot_size) if snapshot_size is not None else f  # type: ignore[assignment]
+        with reader:
```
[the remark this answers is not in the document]
vikramlc-cognite: No longer valid due to the change in the source for the PR.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -100,6 +118,68 @@ def _build_candidate_files(
...
+            cdf_client.files.upload_bytes(
+                content=reader,
+                name=f"{external_id}.log",
+                external_id=external_id,
+                mime_type="text/plain",
+                overwrite=True,
+            )
```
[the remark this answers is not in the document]
vikramlc-cognite: We can remove the code smell and use with open(..) approach which is much more cleaner. Added this change.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -100,6 +117,67 @@ def _build_candidate_files(
...
+        return _FileUploadResult(
+            log_date=candidate.log_date,
+            file_external_id=external_id,
+            status="skipped_too_large",
+            size_bytes=actual_size,
+        )
+
+    try:
+        with open(candidate.path, "rb") as f:
+            reader: BinaryIO = BoundedReader(f, snapshot_size) if snapshot_size is not None else f  # type: ignore[assignment]
```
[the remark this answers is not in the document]
Toshad: @vikramlc Seems legitimate

tests/test_unstable/test_log_upload_action.py
```diff
@@ -297,3 +299,159 @@ def test_range_spanning_rotated_and_live_file(tmp_path: Path) -> None:
...
+
+
+def test_file_external_id_naming_convention() -> None:
+    from cognite.extractorutils.unstable.core._log_upload_action import _file_external_id
+
+    assert _file_external_id("my-extractor", date(2026, 6, 1)) == "extractor-logs-my-extractor-2026-06-01"
+
+
+def test_upload_candidate_calls_cdf_upload_for_rotated_file(tmp_path: Path) -> None:
+    from unittest.mock import MagicMock
```
Yaseen-A-Khan: MagicMock duplicate import

tests/test_unstable/test_log_upload_action.py
```diff
@@ -297,3 +299,159 @@ def test_range_spanning_rotated_and_live_file(tmp_path: Path) -> None:
...
+    assert result.size_bytes == path.stat().st_size
+    assert result.file_external_id == "extractor-logs-my-extractor-2026-06-01"
+    mock_client.files.upload_bytes.assert_called_once()
+    _, kwargs = mock_client.files.upload_bytes.call_args
+    assert kwargs["external_id"] == "extractor-logs-my-extractor-2026-06-01"
+    assert kwargs["overwrite"] is True
+
+
+def test_upload_candidate_current_day_uses_bounded_reader(tmp_path: Path) -> None:
+    from unittest.mock import MagicMock
```
Yaseen-A-Khan: MagicMock duplicate import

tests/test_unstable/test_log_upload_action.py
```diff
@@ -297,3 +299,159 @@ def test_range_spanning_rotated_and_live_file(tmp_path: Path) -> None:
...
+
+    assert result.status == "uploaded"
+    assert result.size_bytes == 300
+    _, kwargs = mock_client.files.upload_bytes.call_args
+    assert isinstance(kwargs["content"], BoundedReader)
+    assert len(kwargs["content"]) == 300
+
+
+def test_upload_candidate_exceeds_max_size_returns_skipped_too_large(tmp_path: Path) -> None:
+    from unittest.mock import MagicMock, patch
```
Yaseen-A-Khan: MagicMock duplicate import

tests/test_unstable/test_log_upload_action.py
```diff
@@ -297,3 +299,159 @@ def test_range_spanning_rotated_and_live_file(tmp_path: Path) -> None:
...
+    with patch("pathlib.Path.stat", return_value=mock_stat_result):
+        result = _upload_candidate(candidate, "my-extractor", mock_client, snapshot_size=None)
+
+    assert result.status == "skipped_too_large"
+    assert result.size_bytes == oversized
+    mock_client.files.upload_bytes.assert_not_called()
+
+
+def test_upload_candidate_cdf_error_returns_failed(tmp_path: Path) -> None:
+    from unittest.mock import MagicMock
```
Yaseen-A-Khan: MagicMock duplicate import

cognite/extractorutils/unstable/core/_log_upload_action.py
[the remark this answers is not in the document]
Yaseen-A-Khan: @vikramlc-cognite along with that, if the `snapshot_size` is None then raw `f` is used.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,78 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
     _logger.info(
         "fetch_logs: %d candidate file(s) for %s to %s; %d date(s) skipped",
         len(candidates),
         start_date,
         end_date,
-        len(skipped),
+        len(skipped_dates),
+    )
+
+    # Snapshot the current-day file size BEFORE spawning upload threads.
```
Yaseen-A-Khan:
``` suggestion
    # Snapshot the current-day file size BEFORE starting uploads.
```
[nit] Can you verify if it is upload threads, because according to Line: 253 it looks like a list comprehension.

Yaseen-A-Khan (changes requested):
I have a few nits and comments, please review them

cognite/extractorutils/unstable/core/_log_upload_action.py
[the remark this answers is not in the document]
vikramlc-cognite: Valid. Added this change.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,78 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+        {
+            "date": str(skipped_date),
+            "file_external_id": _file_external_id(integration_external_id, skipped_date),
+            "status": "skipped",
+        }
+        for skipped_date in sorted(skipped_dates)
+    )
+    files_list.sort(key=lambda e: e["date"])
+
+    total_skipped = len(skipped_dates) + counts["skipped_too_large"]
```
Yaseen-A-Khan: can we have seperate metadata keys like `skipped_missing_files `and `skipped_too_large_files`, so Odin can write precise alerts without having to parse the full files JSON.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -28,6 +36,15 @@ class LogFileCandidate:
     is_current: bool  # True when path is the live (unrotated) file.log
 
 
+@dataclass
+class _FileUploadResult:
+    log_date: date
+    file_external_id: str
+    status: Literal["uploaded", "skipped_too_large", "failed"]
+    size_bytes: int = 0
```
Yaseen-A-Khan:
```suggestion
    size_bytes: Optional[int] = None
```
Changing `size_bytes` to `Optional[int] = None` with an `is not None` guard makes this self-documenting.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,78 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+            _logger.info(
+                "fetch_logs: current-day snapshot %d bytes (%s)",
+                snapshot_size,
+                current_candidate.path.name,
+            )
+        except OSError as e:
+            _logger.warning("fetch_logs: could not snapshot current-day file size — %s", e)
+
+    integration_external_id = ctx._extractor.connection_config.integration.external_id
+    cdf_client = ctx._extractor.cognite_client
```
Yaseen-A-Khan: Since `ActionContext` is now a user-facing extension point (the `report_progress` docstring says it's "intentionally general"), surfacing `ctx.cdf_client` and `ctx.integration_external_id` as explicit properties would close the API surface. Right now `fetch_logs_action` reaching through `ctx._extractor` sets a pattern that user-authored actions would likely copy.

cognite/extractorutils/unstable/core/_bounded_reader.py
```diff
@@ -34,6 +34,14 @@ def __len__(self) -> int:
     def tell(self) -> int:
         return self._size - self._remaining
 
+    def seek(self, offset: int, whence: int = 0) -> int:
+        pos = self._stream.seek(offset, whence)
+        self._remaining = max(0, self._size - pos)
+        return pos
```
Yaseen-A-Khan, line 43: Any client that calls seek(0, 2) to probe file size before retrying an upload gets a broken state.
Either raise `io.UnsupportedOperation` for whence=2, or clamp `pos`to `[0, _size]` before computing `_remaining.`

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,78 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+        files_list.append(entry)
+    files_list.extend(
+        {
+            "date": str(skipped_date),
+            "file_external_id": _file_external_id(integration_external_id, skipped_date),
+            "status": "skipped",
+        }
+        for skipped_date in sorted(skipped_dates)
+    )
+    files_list.sort(key=lambda e: e["date"])
```
Yaseen-A-Khan: Remove `sorted() `from the extend — the outer `.sort` covers it.

Yaseen-A-Khan (commented):
I have some comments, please take a look at them

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: size_bytes = 0 is unambiguous here because _build_candidate_files already filters out empty files before they become candidates, so a _FileUploadResult with size_bytes=0 always means "size not recorded", never "file is actually empty". The if r.size_bytes: guard works correctly on that basis. Since _FileUploadResult is a private dataclass used only within this file, switching to None would just add is not None checks throughout for no real gain.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Good point. It's better to add public properties on ActionContext to avoid access to private internals. Fixed.

cognite/extractorutils/unstable/core/_log_upload_action.py
Yaseen-A-Khan: can you verify, if In case of TOCTOU race, the size might become zero because snapshot_size=0, then it might not work as expected.

cognite/extractorutils/unstable/core/actions.py
```diff
@@ -61,6 +75,11 @@ def _new_error(
             task_name=task_name if task_name is not None else self._action.name,
         )
 
+    def set_result(self, message: str, *, metadata: dict[str, str] | None = None) -> None:
+        """Record the result for a successful action completion."""
+        self._result_message = message
+        self._result_metadata = metadata
```
Yaseen-A-Khan, line 86: A second call to `ctx.set_result()` silently discards the first message and metadata with no log or error.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,77 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+    if current_candidate is not None:
+        try:
+            snapshot_size = current_candidate.path.stat().st_size
+            _logger.info(
+                "fetch_logs: current-day snapshot %d bytes (%s)",
+                snapshot_size,
+                current_candidate.path.name,
+            )
+        except OSError as e:
+            _logger.warning("fetch_logs: could not snapshot current-day file size — %s", e)
```
Yaseen-A-Khan: f the snapshot `stat()` fails here, `snapshot_size` stays None, and `_upload_candidate `falls back to uploading the live file via raw `f` — with no `BoundedReader` cap. That reintroduces the exact race this PR set out to fix: the logger can append bytes between `open()` and `upload_bytes()`, and without a `Content-Length` ceiling that can raise `LocalProtocolError` mid-upload.
I think we should make a change for that and add a test that stubs `Path.stat` to raise on the snapshot call specifically (not the candidate-building stat) to cover this path.

Yaseen-A-Khan (changes requested):
I have some comments, please take a look at them

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -100,6 +117,67 @@ def _build_candidate_files(
...
+        )
+        return _FileUploadResult(
+            log_date=candidate.log_date,
+            file_external_id=external_id,
+            status="skipped_too_large",
+            size_bytes=actual_size,
+        )
+
+    try:
+        with open(candidate.path, "rb") as f:
```
Toshad, line 159: Reopeneing here but bounding it to snapshot_size from above means that we'll miss any line written between these two calls.
What's even the purpose of putting snapshot_size as bound instead of MAX_FILE_SIZE_BYTES ?

Toshad: Nvm, got my answer from a previous diff

cognite/extractorutils/unstable/core/_log_upload_action.py
Toshad: nit: Why not use BoundedReader for all of them?

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,90 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+
+    integration_external_id = ctx.integration_external_id
+    cdf_client = ctx.cdf_client
+
+    # Snapshot the current-day file size BEFORE starting uploads.
+    # This gives a fixed read ceiling for BoundedReader — bytes written after this
+    # point are excluded from the upload, preventing Content-Length mismatches.
+    snapshot_size: int | None = None
+    snapshot_failed_result: _FileUploadResult | None = None
+    current_candidate = next((c for c in candidates if c.is_current), None)
```
Toshad: You're unnecessarily calculating snapshot_size here and then using it as a proxy for candidate.is_current in _upload_candidate().
Just do this check in upload_candidate, and calculate size once. Doing this check in _upload_candidate() instead will capture more lines overall and also keep the code cleaner.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -100,6 +117,67 @@ def _build_candidate_files(
...
+
+    try:
+        with open(candidate.path, "rb") as f:
+            reader: BinaryIO = BoundedReader(f, snapshot_size) if snapshot_size is not None else f  # type: ignore[assignment]
+            cdf_client.files.upload_bytes(
+                content=reader,
+                name=f"{external_id}.log",
+                external_id=external_id,
+                mime_type="text/plain",
+                overwrite=True,
```
Toshad, line 166: This will cause a lot of excessive uploading, especially if clients have workflows. Maybe we should do a hash_check or take a parameter while triggering action.
It can be handled in a later PR, just calling it out.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,90 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+    for r in upload_results:
+        entry: dict[str, str] = {
+            "date": str(r.log_date),
+            "file_external_id": r.file_external_id,
+            "status": r.status,
+        }
+        if r.size_bytes is not None:
+            entry["size_bytes"] = str(r.size_bytes)
+        if r.error:
+            entry["error"] = r.error
```
Toshad, line 258: Do we need these ifs? Can't we upload a dict which has `entry["size_bytes"] = None` or `entry["error"] = None`?

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,90 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+        for skipped_date in skipped_dates
+    )
+    files_list.sort(key=lambda e: e["date"])
+
+    ctx.set_result(
+        f"{counts['uploaded']} of {num_days} log files uploaded to CDF Files",
+        metadata={
+            "total_files": str(num_days),
+            "uploaded_files": str(counts["uploaded"]),
+            "skipped_missing_files": str(len(skipped_dates)),
```
Toshad: nit: `missing files` since extractor didn't skip them deliberately.

Toshad (changes requested):

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Valid. Renamed.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: The dict type is dict[str, str] - None isn't a valid value. More importantly, this JSON is stored as CDF metadata. Including "size_bytes": null and "error": null on every entry (most of which will be uploaded files with no error) adds noise to every record. The conditional guards are intentional: only include a key when it has a meaningful value.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Agree, worth tracking, but not a blocker for this PR.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: The snapshot is computed before the upload loop intentionally, it captures the file size at a single fixed point in time before any uploads begin. If the stat were inside **_upload_candidate**, the snapshot would be taken at upload time, which is later (more bytes may have been appended) and breaks the "point-in-time ceiling" guarantee. If uploads are ever parallelised, each worker would stat independently. The current design is correct.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: The conditional **BoundedReader(f, snapshot_size) if snapshot_size is not None else f** exists precisely because BoundedReader only makes sense for the current-day file. Rotated files are closed and frozen, the logger can't append to them, so there's no race between open() and upload_bytes().
Wrapping a rotated file in BoundedReader would require computing its size just to pass it in, adding pointless overhead for no safety benefit.

cognite/extractorutils/unstable/core/_log_upload_action.py
Toshad, line 63: Why only the first? Are mulitple file handlers actually allowed? Docs seem to disagree?

vikramlc-cognite: Valid. Fixed it.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -100,6 +122,67 @@ def _build_candidate_files(
...
+        return _FileUploadResult(
+            log_date=candidate.log_date,
+            file_external_id=external_id,
+            status="skipped_too_large",
+            size_bytes=actual_size,
+        )
+
+    try:
+        with open(candidate.path, "rb") as f:
+            reader: BinaryIO = BoundedReader(f, snapshot_size) if snapshot_size is not None else f  # type: ignore[assignment]
```
Toshad: My original point about using snapshot_size as is_current is still valid. Use `candidate.is_current` instead.

cognite/extractorutils/unstable/core/_log_upload_action.py
Toshad: Why does the "point-in-time ceiling" have to be across files?
Even if uploads are parallelized, each thread would only read one file anyway? And only 1 file should be current?

In any case, it still doesn't explain why we need a separate error string for stat() on current file vs other files, and why we can't use BoundedReader for all the files.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +227,90 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+                "fetch_logs: could not snapshot current-day file size — %s; skipping to avoid uncapped upload",
+                e,
+            )
+            candidates = [c for c in candidates if not c.is_current]
+            snapshot_failed_result = _FileUploadResult(
+                log_date=current_candidate.log_date,
+                file_external_id=_file_external_id(integration_external_id, current_candidate.log_date),
+                status="failed",
+                error=f"could not read file size for bounded upload: {e}",
+            )
```
Toshad: IMO this piece of code shouldn't exist as there are simpler ways to implement this:
1. Calculate sizes upfront for all files and pass them to upload_candidate when iterating.
2. Run upload_candidate on current file separately first and then on the other files.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -100,6 +122,67 @@ def _build_candidate_files(
...
 
+def _file_external_id(integration_external_id: str, log_date: date) -> str:
+    return f"extractor-logs-{integration_external_id}-{log_date.isoformat()}"
+
+
+def _upload_candidate(
+    candidate: LogFileCandidate,
+    integration_external_id: str,
+    cdf_client: CogniteClient,
+    snapshot_size: int | None,
```
Toshad: nit: Name `snapshot_size` something better if you're keeping it after all. Maybe `max_read_size`.

cognite/extractorutils/unstable/core/_log_upload_action.py
Toshad: Could you create a JIRA if not created?

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: **snapshot_size** is the better name.

The variable serves two roles in _upload_candidate:
1. `actual_size = snapshot_size if snapshot_size is not None else` - it's the reported size of the file
2. `BoundedReader(f, snapshot_size)` - it's the read ceiling

**snapshot_size** fits both uses: it's the size measured at snapshot time, which both represents the file's reported size and caps the read.
**max_read_size** only describes use point 2 and actively conflicts with use point 1 (calling it "**max_read_size**" when you store it in actual_size and log it as the file's byte count is confusing). Also, **max_read_size** sounds like a static policy constant (like **MAX_FILE_SIZE_BYTES**), suggesting a configured limit rather than a measured value.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: **Option 1 – "Calculate sizes upfront for all files":**
- Rotated files are frozen. The OS log rotation has closed/renamed them; nothing will ever append to them. They do not need a **BoundedReader** cap.
- Statting all rotated files just to pass a redundant size into **BoundedReader** adds N extra syscalls for zero safety benefit.
- You'd still need the special pre-loop timing for the current file (the whole point is to snapshot BEFORE uploads start). So you'd have the same complexity plus overhead.

**Option 2 – "Run current file first, then the others":**
- This splits one uniform loop into two passes (current + rotated), meaning separate error handling in both, and losing the clean list comprehension.
- The **snapshot_failed_result** pattern is actually cleaner than embedding a special case mid-upload.

The current design is the correct separation of concerns: snapshot once at a single point in time before the loop, pass it only where needed (the current file), keep the upload loop uniform.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: It's basically the same check. I have used **snapshot_size** to keep it consistent with the checks in **_upload_candidate** method. I don't think this is a blocker for this PR.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Why can't we use BoundedReader for all files?
BoundedReader was specifically written to handle the current day file handling. Refer [here](https://cognitedata.atlassian.net/wiki/spaces/~7120207333078ba8a4415f98a7c58c2f4877a8/pages/6099992944/Design+Doc+Bulk+Log+Upload+Action+all+extractors+support#Current-Day-File-Handling).
Look at how **_upload_candidate** is called in the loop:

```
_upload_candidate(
    candidate,
    integration_external_id,
    cdf_client,
    snapshot_size if candidate.is_current else None,  # <-- None for rotated files
)
```

And inside **_upload_candidate**:

`reader: BinaryIO = BoundedReader(f, snapshot_size) if snapshot_size is not None else f`

So for rotated files, snapshot_size=None is passed → raw f is used. For the current-day file, snapshot_size=<pre-upload-size> is passed → BoundedReader(f, snapshot_size) is used.

The reason it's only needed for the live file:
- Rotated files are frozen. Once log rotation renames file.log → file.log.2026-07-02, the logging handler writes to a new file.log. The rotated file is sealed - its size will never change between open() and upload_bytes(). No race, no Content-Length mismatch possible.
- Current-day file.log is live. The logging handler keeps appending to it while the upload is happening. Without BoundedReader, the HTTP client would read bytes that were written after the upload started, causing a Content-Length mismatch (LocalProtocolError). The snapshot captures the size before the upload begins, and BoundedReader enforces that ceiling.

Using **BoundedReader** for rotated files would require an extra stat() per file just to construct it. If there are 7 rotated files in the date range, that's 7 extra syscalls. Since **_FileUploadResult** already records **size_bytes**, you'd need to stat-then-bound, adding overhead for a scenario that doesn't happen in standard Python logging setups.

The current design makes a deliberate bet: rotated files are frozen, raw f is safe. That's a reasonable bet for the intended use case (standard **TimedRotatingFileHandler**). If you wanted to be maximally defensive, you could use **BoundedReader** for all files - but you'd pay the extra stat calls and the code would need to explain why non-current files also need capping, which would actually be confusing.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -144,11 +222,90 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
+            _logger.warning(
+                "fetch_logs: could not snapshot current-day file size — %s; skipping to avoid uncapped upload",
+                e,
+            )
+            candidates = [c for c in candidates if not c.is_current]
+            snapshot_failed_result = _FileUploadResult(
+                log_date=current_candidate.log_date,
+                file_external_id=_file_external_id(integration_external_id, current_candidate.log_date),
+                status="failed",
+                error=f"could not read file size for bounded upload: {e}",
```
Toshad: Why we need a separate error string for stat() on current file vs other files?

cognite/extractorutils/unstable/core/_log_upload_action.py
Toshad: Option 1: Your points against don't make sense since _upload_candidate already calls stat() on all files (line 134). This stat() is also unavoidable IMO.

Even if that wasn't the case, the stat() call's overhead is negligible and therefore irrelevant when compared to uploading large files. 

Option2:
You misunderstand the whole thing. I mean call _upload_candidate(current_file) then call _upload_candidate(other files). This still ensures the stat happens before any upload begins.

`losing the clean list comprehension.` I don't think the existing code has clean comprehension.

`The snapshot_failed_result pattern is actually cleaner than embedding a special case mid-upload.` The existing code has a special case(if condition) before the upload, during the uploads(BoundedReader), and after the uploads.

cognite/extractorutils/unstable/core/_log_upload_action.py
Toshad: You didn't answer my first question: Why does the "point-in-time ceiling" have to be across files?.
I understand BoundedReader is not _needed_, see below for other objection.

cognite/extractorutils/unstable/core/_log_upload_action.py
Toshad: It's not the correct condition though.

Toshad (changes requested):

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: You're right, it doesn't need to be. The pre-loop snapshot block has been removed entirely. stat() now happens inside **_upload_candidate** for each file, right before open(), the TOCTOU window is identical since BoundedReader still caps reads to the stat-time size regardless of where the stat is called.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: Fixed. **_upload_candidate** no longer takes a snapshot_size parameter, it always calls stat() internally and uses **candidate.is_current** directly to gate **BoundedReader**. The condition is now explicit and self-documenting.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: On deeper look, you were right on both counts. Good find. 
Option 1: _upload_candidate was already calling stat() for rotated files, so the "extra syscalls" argument was wrong. 
Option 2: taken. The entire pre-loop snapshot block (snapshot_size, snapshot_failed_result, current_candidate) has been removed. **_upload_candidate** is now self-contained, it stats and uploads each file uniformly, using **candidate.is_current** to decide whether to wrap with **BoundedReader**. The upload loop is now 3 lines.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: No longer the case. With the pre-loop snapshot removed, all stat() failures go through the same code path inside _upload_candidate and produce a consistent str(e) error. The special **could not read file size for bounded upload** message is gone.

Toshad (dismissed):

thorkildcognite:
Requested changes -> sent it back to the team

Yaseen-A-Khan (dismissed):
LGTM

tuanng-cognite:
One gap: seek()/seekable() are only tested by calling them directly (test_seek_rewinds_for_retry, test_seek_relative, etc.) — there's no test that drives them through an actual retry cycle the way requests/httpx would (read partway → transport error → seek(0) → re-read → re-declare Content-Length).
  And in test_log_upload_action.py, cdf_client is a MagicMock(), so upload_bytes never actually reads from the BoundedReader — we're only asserting it's the right type/length, not that it behaves correctly when actually consumed by an HTTP client.

  Could we add a test that simulates a retry against BoundedReader — e.g. a fake transport/session that reads N bytes, raises, then calls .seek(0) and re-reads — and asserts the second attempt reproduces the same declared length and byte content? That's the scenario this class exists to make safe, and right now it's the one path we're not actually proving works end-to-end.

vikramlc-cognite:
> One gap: seek()/seekable() are only tested by calling them directly (test_seek_rewinds_for_retry, test_seek_relative, etc.) — there's no test that drives them through an actual retry cycle the way requests/httpx would (read partway → transport error → seek(0) → re-read → re-declare Content-Length). And in test_log_upload_action.py, cdf_client is a MagicMock(), so upload_bytes never actually reads from the BoundedReader — we're only asserting it's the right type/length, not that it behaves correctly when actually consumed by an HTTP client.
> 
> Could we add a test that simulates a retry against BoundedReader — e.g. a fake transport/session that reads N bytes, raises, then calls .seek(0) and re-reads — and asserts the second attempt reproduces the same declared length and byte content? That's the scenario this class exists to make safe, and right now it's the one path we're not actually proving works end-to-end.

Good catch. Added **test_retry_cycle_preserves_content_length_and_content** to **test_bounded_reader.py** which simulates the exact retry scenario you described: **super_len()** is called to declare **Content-Length**, then a partial read happens (simulating a mid-upload transport failure), then seek(0) is called, and the second attempt verifies both that super_len() returns the same value and that the full byte content is reproduced correctly.

On the **test_log_upload_action.py** point - agreed that **upload_bytes** never actually consumes the reader there since **cdf_client** is a **MagicMock**. That test covers the orchestration layer rather than the read behaviour itself. The retry correctness is now proven in the **BoundedReader** unit tests directly, which felt like the right place to own that.

tuanng-cognite:
Good progress on test_retry_cycle_preserves_content_length_and_content, but it still doesn't close the gap — it calls read()/seek()/super_len() directly in sequence, same style as the existing test_seek_rewinds_for_retry, just merged into one function. Even a "fake transport" wrapper doesn't really change that if the test itself still decides when to call .seek().

To actually prove the contract, the seek needs to be called by code we don't control — e.g. real urllib3, which is what requests/httpx retry on top of

```
 def test_seek_driven_by_real_urllib3_retry(tmp_path: Path) -> None:
      """
      Unlike a hand-written retry loop, this test never calls reader.seek()
      itself — urllib3 (the library requests/httpx sit on top of) calls it
      internally via rewind_body() when it retries a failed connection.
      This proves BoundedReader satisfies the real contract, not just the
      contract as the test imagines it.
      """
      content = b"hello world retry test data"
      path = _make_file(tmp_path, content)
      attempts = {"count": 0}

      class FlakyConnection:
          is_closed = False
          is_verified = True
          proxy_is_verified = None

          def __init__(self, *a: object, **kw: object) -> None:
              pass

          def request(self, method: str, url: str, body: object = None, **kw: object) -> None:
              attempts["count"] += 1
              if attempts["count"] == 1:
                  body.read(10)  # partial read before the simulated drop
                  raise urllib3.exceptions.ProtocolError("simulated transport failure")
              attempts["second_read"] = body.read()  # urllib3 already rewound body for us

          def getresponse(self) -> urllib3.HTTPResponse:
              return urllib3.HTTPResponse(body=b"", status=200, preload_content=False)

          def close(self) -> None:
              pass

      with open(path, "rb") as f:
          reader = BoundedReader(f, len(content))
          pool = HTTPConnectionPool("example.invalid")
          pool.ConnectionCls = FlakyConnection
          pool.urlopen("POST", "/upload", body=reader, retries=urllib3.Retry(total=1, allowed_methods=None))

      assert attempts["count"] == 2
      assert attempts["second_read"] == content

```
 
Separately, on test_log_upload_action.py,  _upload_candidate has no retry logic of its own (that lives inside the real SDK's HTTP stack, invisible behind MagicMock()), so we can't test a retry there. We can at least make upload_bytes actually read the reader instead of just checking type/length

```
  def test_upload_candidate_bounded_reader_content_is_actually_consumable(tmp_path: Path) -> None:
      path = tmp_path / "extractor.log"
      content = b"log line " * 40
      path.write_bytes(content)
      candidate = LogFileCandidate(log_date=_PAST_TODAY, path=path, is_current=True)
      consumed = {}

      def upload_bytes(*, content: BoundedReader, **kwargs: object) -> None:
          consumed["length"] = len(content)
          consumed["bytes"] = content.read()

      mock_client = MagicMock()
      mock_client.files.upload_bytes.side_effect = upload_bytes

      result = _upload_candidate(candidate, "my-extractor", mock_client)

      assert result.status == "uploaded"
      assert consumed["length"] == len(content)
      assert consumed["bytes"] == content
```

Together these cover both points from my original comment: the actual retry-cycle contract (proven against a fake transport, since a real one isn't practical to mock at this layer) and real consumption of the reader through the orchestration code path (not just isinstance/len checks).

vikramlc-cognite:
> Good progress on test_retry_cycle_preserves_content_length_and_content, but it still doesn't close the gap — it calls read()/seek()/super_len() directly in sequence, same style as the existing test_seek_rewinds_for_retry, just merged into one function. Even a "fake transport" wrapper doesn't really change that if the test itself still decides when to call .seek().
> 
> To actually prove the contract, the seek needs to be called by code we don't control — e.g. real urllib3, which is what requests/httpx retry on top of
> 
> ```
>  def test_seek_driven_by_real_urllib3_retry(tmp_path: Path) -> None:
>       """
>       Unlike a hand-written retry loop, this test never calls reader.seek()
>       itself — urllib3 (the library requests/httpx sit on top of) calls it
>       internally via rewind_body() when it retries a failed connection.
>       This proves BoundedReader satisfies the real contract, not just the
>       contract as the test imagines it.
>       """
>       content = b"hello world retry test data"
>       path = _make_file(tmp_path, content)
>       attempts = {"count": 0}
> 
>       class FlakyConnection:
>           is_closed = False
>           is_verified = True
>           proxy_is_verified = None
> 
>           def __init__(self, *a: object, **kw: object) -> None:
>               pass
> 
>           def request(self, method: str, url: str, body: object = None, **kw: object) -> None:
>               attempts["count"] += 1
>               if attempts["count"] == 1:
>                   body.read(10)  # partial read before the simulated drop
>                   raise urllib3.exceptions.ProtocolError("simulated transport failure")
>               attempts["second_read"] = body.read()  # urllib3 already rewound body for us
> 
>           def getresponse(self) -> urllib3.HTTPResponse:
>               return urllib3.HTTPResponse(body=b"", status=200, preload_content=False)
> 
>           def close(self) -> None:
>               pass
> 
>       with open(path, "rb") as f:
>           reader = BoundedReader(f, len(content))
>           pool = HTTPConnectionPool("example.invalid")
>           pool.ConnectionCls = FlakyConnection
>           pool.urlopen("POST", "/upload", body=reader, retries=urllib3.Retry(total=1, allowed_methods=None))
> 
>       assert attempts["count"] == 2
>       assert attempts["second_read"] == content
> ```
> 
> Separately, on test_log_upload_action.py, _upload_candidate has no retry logic of its own (that lives inside the real SDK's HTTP stack, invisible behind MagicMock()), so we can't test a retry there. We can at least make upload_bytes actually read the reader instead of just checking type/length
> 
> ```
>   def test_upload_candidate_bounded_reader_content_is_actually_consumable(tmp_path: Path) -> None:
>       path = tmp_path / "extractor.log"
>       content = b"log line " * 40
>       path.write_bytes(content)
>       candidate = LogFileCandidate(log_date=_PAST_TODAY, path=path, is_current=True)
>       consumed = {}
> 
>       def upload_bytes(*, content: BoundedReader, **kwargs: object) -> None:
>           consumed["length"] = len(content)
>           consumed["bytes"] = content.read()
> 
>       mock_client = MagicMock()
>       mock_client.files.upload_bytes.side_effect = upload_bytes
> 
>       result = _upload_candidate(candidate, "my-extractor", mock_client)
> 
>       assert result.status == "uploaded"
>       assert consumed["length"] == len(content)
>       assert consumed["bytes"] == content
> ```
> 
> Together these cover both points from my original comment: the actual retry-cycle contract (proven against a fake transport, since a real one isn't practical to mock at this layer) and real consumption of the reader through the orchestration code path (not just isinstance/len checks).

Thank you. This gives a good starting point for tests I can include for such scenarios in future. 
I have addressed your comments with the tests you provided:
For the retry test, replaced the hand-written seek loop with **test_seek_driven_by_real_urllib3_retry**, the test itself never calls reader.seek(). A FlakyConnection reads 10 bytes then raises ProtocolError on the first attempt, and urllib3's internal rewind_body() calls seek(0) before retrying. The second attempt asserts it gets the full content back.

For the consumption test, added **test_upload_candidate_bounded_reader_content_is_actually_consumable**, upload_bytes now has a real side_effect that calls len(content) and content.read() on the BoundedReader, asserting both the declared length and the byte content match the file.

tuanng-cognite (approved):
🦄

Yaseen-A-Khan (approved):

PR #555, opened 2026-07-14
fix(odin): Fix fetch logs result metadata

vikramlc-cognite:
### Summary
Fixes the fetch_logs action so its result metadata can't exceed CDF's 512-byte-per-value limit, and prevents oversized metadata from silently poisoning the checkin batch.

### Type of change

### What changed
- `cognite/extractorutils/unstable/core/_dto.py`: added **MAX_METADATA_VALUE_BYTES** (512) constant and **oversized_metadata_fields()** helper to detect metadata dict values exceeding CDF's per-value byte limit.
- `cognite/extractorutils/unstable/core/_log_upload_action.py`:
  - **_FileUploadResult** now carries file_id (the numeric CDF file ID returned by upload_bytes), set only when status == "uploaded".
  - Per-file files metadata entries now use the compact numeric id instead of the full file_external_id string, and only include size_bytes for non-uploaded entries - shrinking the payload so it fits within **MAX_METADATA_VALUE_BYTES** even at **MAX_DATE_RANGE_DAYS**.
  - Added a size check before attaching files to the result metadata: if the serialized JSON still exceeds the 512-byte limit, it's omitted (with a log line and a note appended to the result message) rather than risking the whole result being rejected. The aggregate counts (total_files, uploaded_files, etc.) are always included regardless.
- `cognite/extractorutils/unstable/core/base.py`: in **_dispatch_single_action**, both the success and ActionError paths now check **oversized_metadata_fields()** before queuing the checkin update:
  - Success path: if any result metadata field is oversized, the action is reported as failed instead (with an explanatory message) rather than queuing a payload Odin would reject.
  - ActionError path: if **e.result_metadata** has oversized fields, they're dropped and the failure message notes which fields were dropped, instead of sending metadata that would get rejected.
- Tests added/updated in `tests/test_unstable/test_action_dispatch.py` and `tests/test_unstable/test_log_upload_action.py` covering all of the above.

### Why it changed
- Related issue: EDG-678
- Related docs / discussion: Odin's checkin API rejects the entire batch of pending action updates if any single metadata value exceeds 512 bytes ("Metadata values may be at most 512 bytes"). Because checkin bundles all pending updates and requeues the whole batch on rejection, one oversized files value (previously embedding full file external IDs per day) could poison retries indefinitely for **fetch_logs** runs over long date ranges.

- The reasoning in **base.py** for turning an otherwise-successful action into a failed checkin update when its result metadata is oversized - confirm this fail-closed behaviour (vs. silently truncating/dropping just the offending field) is the desired tradeoff.
- The fallback behaviour in **_log_upload_action.py** when files still doesn't fit even after switching to numeric IDs (aggregate counts still reported, files omitted, message annotated).
- Whether omitting file_external_id from per-file entries (replaced by numeric id) is acceptable for any downstream consumers of that metadata field.

### Test evidence
- New/updated unit tests: `test_oversized_result_metadata_fails_action_instead_of_reporting_succeeded`,
`test_oversized_action_error_metadata_drops_metadata_but_keeps_original_message`, `test_max_range_with_uploads_keeps_files_metadata_within_budget`,
`test_files_metadata_omitted_when_still_oversized`.

### Risks and unknowns
- Any external tooling/dashboards that currently parse **file_external_id** out of the files metadata field would need to switch to reconstructing it from id/date, or querying CDF by file ID.

### Rollout and rollback
- No migrations or config changes. Pure behaviour change in metadata reporting, revertible via normal git revert.

### DIFF
--- a/cognite/extractorutils/unstable/core/_log_upload_action.py
+++ b/cognite/extractorutils/unstable/core/_log_upload_action.py
@@ -13,6 +13,7 @@ from cognite.client import CogniteClient
 
 from cognite.extractorutils.unstable.configuration.models import ExtractorConfig, LogFileHandlerConfig
 from cognite.extractorutils.unstable.core._bounded_reader import BoundedReader
+from cognite.extractorutils.unstable.core._dto import MAX_METADATA_VALUE_BYTES
 from cognite.extractorutils.unstable.core.actions import ActionContext, ActionError
 
 _logger = logging.getLogger(__name__)
@@ -41,6 +42,7 @@ class _FileUploadResult:
     log_date: date
     file_external_id: str
     status: Literal["uploaded", "skipped_too_large", "failed"]
+    file_id: int | None = None  # CDF's internal (numeric) file id; set only when status == "uploaded"
     size_bytes: int | None = None
     error: str | None = None
 
@@ -158,7 +160,7 @@ def _upload_candidate(
     try:
         with open(candidate.path, "rb") as f:
             reader: BinaryIO = BoundedReader(f, actual_size) if candidate.is_current else f  # type: ignore[assignment]
-            cdf_client.files.upload_bytes(
+            uploaded = cdf_client.files.upload_bytes(
                 content=reader,
                 name=f"{external_id}.log",
                 external_id=external_id,
@@ -170,6 +172,7 @@ def _upload_candidate(
             log_date=candidate.log_date,
             file_external_id=external_id,
             status="uploaded",
+            file_id=uploaded.id,
             size_bytes=actual_size,
         )
     except Exception as e:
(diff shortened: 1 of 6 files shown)


cognite/extractorutils/unstable/core/base.py
```diff
@@ -676,21 +678,52 @@ def _handle_custom_action(self, action: Action) -> None:
...
         try:
             custom.target(ctx)
-            self._checkin_worker.queue_action_update(
-                ActionUpdate(
-                    external_id=action.external_id,
-                    status=ActionStatus.succeeded,
-                    result_message=ctx._result_message,
-                    result_metadata=ctx._result_metadata,
+            oversized_fields = oversized_metadata_fields(ctx._result_metadata)
+            if oversized_fields:
```
mohammad-arif662, line 683: On the success path, if any metadata field is oversized we drop the whole result_metadata dict and flip the action to failed, even fields that were fine. The `ActionError` branch below does the opposite, it only strips the keys that are oversized and keeps the rest. Is it intentional ?

cognite/extractorutils/unstable/core/base.py
```diff
@@ -676,21 +678,52 @@ def _handle_custom_action(self, action: Action) -> None:
...
+                    result_message=(
+                        str(e)
+                        if not oversized_fields
+                        else (
+                            f"{e} (additionally, metadata field(s) {', '.join(oversized_fields)} exceeded "
+                            f"the {MAX_METADATA_VALUE_BYTES}-byte-per-value limit and were dropped)"
+                        )
+                    ),
+                    result_metadata=(
+                        {k: v for k, v in e.result_metadata.items() if k not in oversized_fields}
```
mohammad-arif662: If every key in `e.result_metadata` is oversized, this becomes `{}` rather than `None`. Does Integration API handles it ?

tests/test_unstable/test_action_dispatch.py
```diff
@@ -184,6 +184,39 @@ def target(ctx: ActionContext) -> None:
     assert failed.result_message == "bad input"
 
 
+def test_oversized_result_metadata_fails_action_instead_of_reporting_succeeded() -> None:
```
mohammad-arif662: This test only has one metadata field, so it can't tell which path it took, success with oversized or exception block, because both give the same result_metadata is None result. may be add one more field {"summary": "ok", "blob": "x" * 600} and assert whether "summary" is there or not, in both flows.

mohammad-arif662 (dismissed):

cognite/extractorutils/unstable/core/base.py
```diff
@@ -676,21 +678,49 @@ def _handle_custom_action(self, action: Action) -> None:
...
+                # The action itself ran to completion — only reporting the full result back to Odin
+                # failed, because a metadata value is too large to send. Fail the action instead of
+                # queuing a payload that Odin would reject (which would otherwise poison the checkin
+                # batch and retry forever, since checkin bundles all pending updates together and
+                # requeues the whole batch on any rejection). Non-oversized fields are still reported.
+                self._checkin_worker.queue_action_update(
+                    ActionUpdate(
+                        external_id=action.external_id,
+                        status=ActionStatus.failed,
+                        result_message=(
```
tuanng-cognite: result_message is constrained to 1000 characters. ', '.join(oversized_fields) -> could go beyond that limit. Constructing ActionUpdate raises a pydantic.ValidationError. We should truncate or cap the number of fields.

cognite/extractorutils/unstable/core/_log_upload_action.py
```diff
@@ -246,37 +249,45 @@ def fetch_logs_action(ctx: ActionContext) -> None:
...
-        entry: dict[str, str] = {
-            "date": str(r.log_date),
-            "file_external_id": r.file_external_id,
-            "status": r.status,
-        }
-        if r.size_bytes is not None:
+        entry: dict[str, str] = {"date": str(r.log_date), "status": r.status}
+        if r.file_id is not None:
+            entry["id"] = str(r.file_id)
+        if r.status != "uploaded" and r.size_bytes is not None:
```
tuanng-cognite, line 258: can you confirm that there is no consumers of this method which relies on file_external_id? This change will remove that field.

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: That's a nice catch. Added truncation.

cognite/extractorutils/unstable/core/_log_upload_action.py
vikramlc-cognite: There are no consumers for **file_external_id** as this is part of a feature which is still in testing.

tuanng-cognite:
:unicorn:

mohammad-arif662 (approved):

PR #560, opened 2026-07-22
feat(odin): Action support in ContinuousTask

vikramlc-cognite:
## Summary
Extends the Start/Stop action mechanism (previously `ScheduledTask`-only) to also cover `ContinuousTask`, so a continuously-running task can now be stopped and restarted on demand via Odin, the same way scheduled tasks already can.

## Type of change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Chore / tooling / CI

## What changed
- `base.py`: widened five `isinstance(t, ScheduledTask)` gates to `isinstance(t, (ScheduledTask, ContinuousTask))` — `add_action`'s and `add_task`'s reserved-name checks, `_get_startup_request`'s `Start`/`Stop` action registration and `action=` bool on the reported task, `_dispatch_single_action`'s name-matching sets, and `_handle_start_task_action`'s task lookup.
- `_run_task_with_token`'s type hint widened from `ScheduledTask` to `ScheduledTask | ContinuousTask` — no logic change, it only ever touched `.name`/`.target`.
- Extracted the continuous-task launch out of `run()` into a new `_launch_continuous_task(task)` method. It now pre-registers the task's cancellation token in `_running_task_tokens` *before* starting the thread (previously the token was created and handed straight to the thread, never tracked), then runs the task through `_run_task_with_token` instead of calling `task.target` directly. `run()` now just calls this once per continuous task.
- No changes needed to `_handle_stop_task_action` (already generic — looks up by name only) or to the automatic-scheduling `match` in `add_task` / the task-classification `match` in `run()` (a `ContinuousTask` still isn't auto-scheduled via `TaskScheduler`, and still launches once at boot — it's just trackable now).
- `fetch_logs` and db-extractor needed no changes: `fetch_logs` was already task-type-agnostic, and db-extractor doesn't currently define any `ContinuousTask`.
- Tests: split the one test that asserted `ContinuousTask` gets *no* actions into a `StartupTask`-only variant, added a test confirming `ContinuousTask` now gets `Start`/`Stop` actions with the correct `ActionType`/`task` fields, and added four tests in `test_action_dispatch.py` covering the actual runtime behavior — token registered before the thread starts (no race), `Stop` cancels a boot-launched instance, `Start` on an already-running instance reports `failed`, and `Start` after `Stop` successfully relaunches it.

## Why it changed
- Related issue: EDG-714
- Related docs / discussion: follow-up to the `fetch_logs` result-metadata fix (#555 / 7.13.1) — while testing that, we found Start/Stop was scoped to `ScheduledTask` only, with no way to stop/restart a `ContinuousTask` on demand.

- The core risk is in `_launch_continuous_task`: the token must be registered in `_running_task_tokens` *before* the thread starts, not inside it, or there's a race where a Stop/Start arriving immediately after boot could miss it. Worth double-checking the lock scoping there.
- `_handle_start_task_action` runs the task synchronously inside its own per-action dispatch thread. For a `ContinuousTask` (which runs indefinitely), this means the action's own status stays `running` until a `Stop` cancels it — this is a deliberate design choice (reuses the existing dispatch-thread-per-action model with no new threading), not an oversight, but worth confirming that's the semantic you want rather than, say, reporting `succeeded` immediately after a fire-and-forget launch.
- Crash-then-restart interaction: `_run_task_with_token`'s `finally` clears the registry entry regardless of how the target exited (including a crash, since `add_task`'s wrapped `run_task` swallows exceptions internally before `RESTART_POLICY` decides whether to restart the whole extractor). That means a `Start` action can immediately relaunch a continuous task that crashed on its own — intentional, but double-check it doesn't fight with `RESTART_POLICY` in a way that's surprising for extractors using `WHEN_CONTINUOUS_TASKS_CRASHES`.

## Test evidence
- `poetry run pytest tests/test_unstable/test_action_dispatch.py tests/test_unstable/test_action_registration.py -q` → 43 passed, run repeatedly (3x) to rule out flakiness in the new threading-based tests.
- `poetry run pytest tests/test_unstable/ -q` → 237 passed, 1 failed, 48 errors — the failure and errors are pre-existing (missing `COGNITE_*` env vars in the local shell), confirmed identical against the unmodified base via `git stash`.
- `poetry run mypy cognite/extractorutils/unstable/core/` → no issues (13 files).
- `poetry run ruff check cognite/extractorutils/unstable/core/` → all checks passed.

## Risks and unknowns
- No test exercises the real `Extractor.run()` entrypoint end-to-end (it also drives startup tasks and blocks on the scheduler/cancellation wait, which makes it awkward to test directly) — coverage is via the extracted `_launch_continuous_task` method instead, which is exercised directly.
- Haven't verified behavior when a `ContinuousTask`'s target doesn't check its `cancellation_token` cooperatively — same pre-existing caveat as `ScheduledTask`, just now more visible since Stop is exposed for this task type too.
- CHANGELOG.md not yet updated for this change.

## Rollout and rollback
- Purely additive to the action-dispatch surface (new `isinstance` branches, one extracted method) — no config or migration required.
Existing extractors using only `ScheduledTask`/`StartupTask` are unaffected. Revert is a plain `git revert` of this commit if needed.

## Checklist
- [x] Self-reviewed the diff
- [ ] Docs updated (or N/A) — CHANGELOG.md entry still outstanding
- [x] Breaking changes called out above and communicated to affected teams — N/A, not a breaking change

### DIFF
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -382,12 +382,13 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             raise ValueError(f"Custom action '{action.name}' is already registered.")
 
         reserved_names = {
-            name for t in self._tasks if isinstance(t, ScheduledTask) for name in (f"Start {t.name}", f"Stop {t.name}")
+            name
+            for t in self._tasks
+            if isinstance(t, (ScheduledTask, ContinuousTask))
+            for name in (f"Start {t.name}", f"Stop {t.name}")
         }
         if action.name in reserved_names:
-            raise ValueError(
-                f"Custom action name '{action.name}' conflicts with an auto-generated scheduled task action."
-            )
+            raise ValueError(f"Custom action name '{action.name}' conflicts with an auto-generated task action.")
 
         self._custom_actions.append(action)
 
@@ -402,7 +403,7 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         available_actions: list[AvailableActionWrite] = []
 
         for t in self._tasks:
-            if isinstance(t, ScheduledTask):
+            if isinstance(t, (ScheduledTask, ContinuousTask)):
                 available_actions.append(
                     AvailableActionWrite(name=f"Start {t.name}", type=ActionType.start_task, task=t.name)
                 )
@@ -422,7 +423,7 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             tasks=[
                 DtoTask(
                     type=TaskType.continuous if isinstance(t, ContinuousTask) else TaskType.batch,
-                    action=isinstance(t, ScheduledTask),
+                    action=isinstance(t, (ScheduledTask, ContinuousTask)),
                     description=t.description,
                     name=t.name,
                 )
@@ -484,7 +485,7 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         if any(t.name == task.name for t in self._tasks):
             raise ValueError(f"Task '{task.name}' is already registered.")
 
-        if isinstance(task, ScheduledTask):
+        if isinstance(task, (ScheduledTask, ContinuousTask)):
             conflict_names = {f"Start {task.name}", f"Stop {task.name}"}
             for action in self._custom_actions:
                 if action.name in conflict_names:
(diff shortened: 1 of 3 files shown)


cognite/extractorutils/unstable/core/base.py
```diff
@@ -558,6 +562,23 @@ def _run_task_with_token(self, task: ScheduledTask, child_token: CancellationTok
...
+        there is no race window where a Stop/Start action could run before this instance is visible.
+        """
+        child_token = self.cancellation_token.create_child_token()
+        with self._running_task_tokens_lock:
+            self._running_task_tokens[task.name] = child_token
+        Thread(
+            name=pascalize(task.name),
+            target=self._run_task_with_token,
+            args=(task, child_token),
+        ).start()
```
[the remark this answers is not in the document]
vikramlc-cognite: Valid point.

Yaseen-A-Khan:
changes look good, just one nit question.
will the CHANGELOG.md get updated as part of this PR?

vikramlc-cognite:
> changes look good, just one nit question. will the CHANGELOG.md get updated as part of this PR?

Will add the changelog and update the version as part of a different PR where I need to fix the task status during extractor shutdown. Jira for the same here: https://cognitedata.atlassian.net/browse/EDG-717

Yaseen-A-Khan (dismissed):
LGTM

cognite/extractorutils/unstable/core/base.py
```diff
@@ -382,12 +382,13 @@ def add_action(self, action: CustomAction) -> None:
             raise ValueError(f"Custom action '{action.name}' is already registered.")
 
         reserved_names = {
-            name for t in self._tasks if isinstance(t, ScheduledTask) for name in (f"Start {t.name}", f"Stop {t.name}")
+            name
+            for t in self._tasks
+            if isinstance(t, (ScheduledTask, ContinuousTask))
```
andersfylling: can you move the tuple of acceptable candidates to a global var / const to avoid drift?

andersfylling (changes requested):

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Valid. Added a global variable for all actionable task types.

andersfylling (approved):

andersfylling:
:unicorn:

Yaseen-A-Khan (approved):

PR #571, opened 2026-08-26
fix(odin): Update stop action status to succeeded instead of canceled

vikramlc-cognite:
## Summary
Fixes stop-task actions incorrectly reporting `Canceled` instead of `Succeeded` when they successfully stop the target task.

## Type of change
- [ ] Documentation update
- [ ] Chore / tooling / CI

## What changed
- `cognite/extractorutils/unstable/core/base.py::_handle_stop_task_action`: report `ActionStatus.succeeded` (was `ActionStatus.canceled`) when the target task's `CancellationToken` is found and cancelled.
- Updated the two affected assertions in `tests/test_unstable/test_action_dispatch.py` (`test_stop_task_action_cancels_child_token_and_reports_succeeded`, `test_stop_action_cancels_boot_launched_continuous_task`) to expect `succeeded`.

## Why it changed
`ActionStatus.canceled` is meant to mean "this action was itself canceled," not "this action successfully performed a cancellation." A stop action that does its job correctly was showing up as aborted in Odin, misleading operators.

- Confirm no downstream consumer (Odin UI/alerting) treats a stop action's terminal status as a signal distinct from `succeeded` in a way that depended on the old `canceled` value.

## Test evidence
- `python -m pytest tests/test_unstable/test_action_dispatch.py -q` → 27 passed.

## Risks and unknowns
- None identified; single status value change, `failed`/token-not-found path unaffected.

## Rollout and rollback
- No flags/migrations. Plain code revert if needed.

## Checklist
- [x] Self-reviewed the diff
- [ ] Docs updated (or N/A) — N/A, internal status semantics, not user-facing docs
- [x] Breaking changes called out above and communicated to affected teams — N/A, not breaking

### DIFF
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -696,7 +696,7 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             return
 
         self._checkin_worker.queue_action_update(
-            ActionUpdate(external_id=action.external_id, status=ActionStatus.canceled)
+            ActionUpdate(external_id=action.external_id, status=ActionStatus.succeeded)
         )
 
     def _handle_custom_action(self, action: Action) -> None:
--- a/tests/test_unstable/test_action_dispatch.py
+++ b/tests/test_unstable/test_action_dispatch.py
@@ -121,7 +121,7 @@ def test_stop_task_action_not_running_reports_failed() -> None:
     assert "not currently running" in (updates[0].result_message or "")
 
 
-def test_stop_task_action_cancels_child_token_and_reports_canceled() -> None:
+def test_stop_task_action_cancels_child_token_and_reports_succeeded() -> None:
     extractor = _make_extractor()
     task_started = Event()
     allow_exit = Event()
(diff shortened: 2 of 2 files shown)


Toshad (approved):

andersfylling:
I don't quite understand this? If I cancelled a task and I would expect it to show up as cancelled?

vikramlc-cognite:
> I don't quite understand this? If I cancelled a task and I would expect it to show up as cancelled?

In this system, a Task (e.g. "Task1") is the actual running job - but a Task itself doesn't carry a **succeeded/failed/canceled** status. Only Actions do. When you click "Stop" on Task1, that's not a status change on Task1 directly - it dispatches a brand-new, separate **Action** called **"Stop Task1" (type stop_task)**.

Meanwhile, **Task1** is actually being run by a different Action - the **"Start Task1" action (type start_task)**, which has been sitting there this whole time, blocked executing Task1's code until it finishes or is interrupted.

So when you cancel Task1, there are two separate Action records involved, each with its own independent status:

**"Start Task1"** - the action that's actually running the task. When "Stop Task1" interrupts it, this action correctly reports canceled - because the thing it was doing (running Task1) really was cut short. This is what you are picturing when you say "I cancelled a task, I'd expect it to show cancelled" - and it's still exactly true after this PR. Nothing about this changes.

**"Stop Task1"** - a separate, short-lived action whose only job is to signal "Start Task1" to stop. This is the one this PR touches. Before the fix, whenever "Stop Task1" successfully delivered that signal, it also reported its own status as canceled - which reads as if the stop request itself was aborted or failed, when actually it did exactly what it was supposed to do. After the fix, "Stop Task1" reports succeeded (the stop request worked), while "Start Task1" - the action that was actually running your task - still reports canceled, unaffected.

So the confusion is: you're thinking of Task1's own outcome (still canceled, unchanged), but the status this PR changes belongs to a different action - the "Stop" request itself, not the thing it stopped.

kornelione (commented):
🦄

PR #572, opened 2026-08-26
fix(odin): Fixed cancelling an in-flight action

vikramlc-cognite:
## Summary
Fixes cancelling an in-flight `start_task`/`custom` action, plus three follow-up correctness fixes surfaced during review: status/message accuracy when oversized metadata coincides with cancellation, and a token-registration leak on early failure.

## Type of change
- [ ] Documentation update
- [ ] Chore / tooling / CI

## What changed
**Core fix — cancel an in-flight action instead of re-dispatching it:**
- `base.py`: added `_running_action_tokens` (keyed by `Action.external_id`, mirroring `_running_task_tokens`), populated by `_handle_start_task_action`/`_handle_custom_action`.
- `base.py`: `_dispatch_single_action` now checks `action.status` up front — a `cancel_pending` re-delivery (Odin's signal that a user cancelled an already-dispatched action) cancels the tracked token by `external_id` and returns, instead of re-running the handler.
- `base.py`: `_handle_custom_action` reports `ActionStatus.canceled` (not `succeeded`) if the action's token was cancelled before the target returned.
- `actions.py`: `ActionContext` gains a required `cancellation_token: CancellationToken`, so custom actions can cooperatively check `ctx.cancellation_token.is_cancelled`/`.wait(...)`.

**Follow-up fixes (from review):**
- `base.py`: **behavioral change** — the oversized-metadata branch in `_handle_custom_action` used to *always* report `ActionStatus.failed`, regardless of whether the action itself succeeded. It now reports the action's real outcome (`succeeded`/`canceled`) instead — metadata truncation is reflected only in the message and dropped fields, not the status. Any consumer (dashboards, alerting, automation) relying on "oversized metadata → failed" as a signal will see a different status for these actions after this PR. Message wording also no longer claims "completed successfully" when the action was actually cancelled.
- `base.py`: token registration (`_running_action_tokens`, and `_running_task_tokens` for start_task) now happens inside the same `try`/`finally` that cleans it up, in both `_handle_start_task_action` and `_handle_custom_action` — previously a failure between registration and the actual work (e.g. constructing the "running" `ActionUpdate`) would leave the task/action stuck "running" forever with no error ever reported.
- `base.py`: extracted the repeated "release token if still owned" lock/get/pop idiom (5 call sites) into a shared `_release_if_owned` helper.

**Tests:** updated all `ActionContext(...)` call sites (`test_actions.py`, `test_log_upload_action.py`) to pass a token; added 8 new tests to `test_action_dispatch.py` covering cancel-pending re-delivery (custom and start_task actions), cooperative cancellation reporting, oversized-metadata status/wording under cancellation, and registration cleanup on early failure.

## Why it changed
Odin propagates a cancel of a running action by flipping it to `cancel_pending` and re-sending the same `external_id` on the next checkin. The SDK never read `Action.status`, so this was treated as a fresh dispatch — re-running custom action side effects a second time, or spuriously failing an already-running start_task action. The review follow-ups close two adjacent gaps in the same area: status/message accuracy was still wrong in one corner case (oversized metadata + cancellation), and the new token-tracking dicts had the same "register before try" fragility as existing code nearby.

- **Behavioral change:** oversized-metadata custom actions now report `succeeded`/`canceled` instead of always `failed` — see "Follow-up fixes" above. Flag if any downstream consumer depends on the old always-`failed` behavior for this case.
- `ActionContext.cancellation_token` is a new required constructor argument — any external code constructing `ActionContext` directly needs to pass one.
- Cancellation remains cooperative — a custom action that never checks the token still runs to completion; this PR only prevents double-execution and misreported status, not forced interruption.
- `_release_if_owned` is a pure mechanical refactor (verified via test diff against pre-refactor code) — no behavior change intended.

## Test evidence
- `pytest tests/test_unstable/test_action_dispatch.py tests/test_unstable/test_actions.py tests/test_unstable/test_log_upload_action.py -q` → 96 passed.
- `pytest tests/test_unstable/ -q` → 248 passed (48 pre-existing, unrelated errors from missing local env vars; 0 new failures).
- `mypy` / `ruff check` / `ruff format` → clean.

## Risks and unknowns
- The oversized-metadata status change (see above) is the one behavior change with real external-visibility risk in this PR — worth confirming no dashboard/alert depends on the old `failed`-always behavior before merge.
- Custom actions that don't check the cancellation token get no benefit beyond "no longer double-executed" — inherent cooperative-cancellation limitation, not addressed here.
- `stop_task` action behavior is unchanged (completes near-instantly; cancel-pending race window is negligible there).

## Rollout and rollback
No flags/migrations. `ActionContext`'s new required parameter is the only signature change; plain code revert if needed.

## Checklist
- [x] Self-reviewed the diff
- [x] Tests added or updated
- [ ] Docs updated (or N/A) — N/A, internal SDK behavior; documented via docstrings
- [x] Breaking changes called out above — `ActionContext` now requires `cancellation_token`; oversized-metadata status behavior change called out under "What to focus on during review"

### DIFF
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -203,6 +203,11 @@ class Extractor(Generic[ConfigType], CogniteLogger):
         self._custom_actions: list[CustomAction] = []
         self._running_task_tokens: dict[str, CancellationToken] = {}
         self._running_task_tokens_lock = RLock()
+        # Keyed by Action.external_id (not task name) so a `cancel_pending` re-delivery of an
+        # in-flight start_task/custom action can be cancelled instead of re-dispatched. Populated by
+        # _handle_start_task_action/_handle_custom_action, consulted by _dispatch_single_action.
+        self._running_action_tokens: dict[str, CancellationToken] = {}
+        self._running_action_tokens_lock = RLock()
         self._start_time: datetime
 
         self.metrics: BaseMetrics = self._load_metrics(config.metrics_class)
@@ -612,6 +617,16 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             ).start()
 
     def _dispatch_single_action(self, action: Action) -> None:
+        if action.status == ActionStatus.cancel_pending:
+            # Odin re-delivers an already-dispatched action once a user cancels it, with its status
+            # flipped to cancel_pending. This is not a fresh invocation: cancel the token tracking the
+            # in-flight run (if we have one) and stop, rather than re-running the action's handler.
+            with self._running_action_tokens_lock:
+                token = self._running_action_tokens.get(action.external_id)
+            if token is not None:
+                token.cancel()
+            return
+
         actionable_tasks = [t for t in self._tasks if isinstance(t, ACTIONABLE_TASK_TYPES)]
         scheduled_start_names = {f"Start {t.name}" for t in actionable_tasks}
         scheduled_stop_names = {f"Stop {t.name}" for t in actionable_tasks}
@@ -661,6 +676,9 @@ class Extractor(Generic[ConfigType], CogniteLogger):
                 return
             self._running_task_tokens[task_name] = child_token
 
+        with self._running_action_tokens_lock:
+            self._running_action_tokens[action.external_id] = child_token
+
         self._checkin_worker.queue_action_update(
             ActionUpdate(external_id=action.external_id, status=ActionStatus.running)
         )
@@ -677,6 +695,10 @@ class Extractor(Generic[ConfigType], CogniteLogger):
                     result_message=str(e),
                 )
             )
+        finally:
+            with self._running_action_tokens_lock:
+                if self._running_action_tokens.get(action.external_id) is child_token:
+                    self._running_action_tokens.pop(action.external_id, None)
 
     def _handle_stop_task_action(self, action: Action) -> None:
         task_name = action.action_name[len("Stop ") :]
@@ -711,6 +733,10 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             )
             return
 
+        action_token = self.cancellation_token.create_child_token()
+        with self._running_action_tokens_lock:
+            self._running_action_tokens[action.external_id] = action_token
+
         self._checkin_worker.queue_action_update(
             ActionUpdate(external_id=action.external_id, status=ActionStatus.running)
         )
@@ -720,11 +746,13 @@ class Extractor(Generic[ConfigType], CogniteLogger):
             extractor=self,
             external_id=action.external_id,
             call_metadata=action.call_metadata,
+            cancellation_token=action_token,
         )
 
         try:
             custom.target(ctx)
             filtered_metadata, oversized_fields = drop_oversized_metadata_fields(ctx._result_metadata)
+            completed_status = ActionStatus.canceled if action_token.is_cancelled else ActionStatus.succeeded
             if oversized_fields:
                 # The action itself ran to completion — only reporting the full result back to Odin
                 # failed, because a metadata value is too large to send. Fail the action instead of
@@ -747,7 +775,7 @@ class Extractor(Generic[ConfigType], CogniteLogger):
                 self._checkin_worker.queue_action_update(
                     ActionUpdate(
                         external_id=action.external_id,
-                        status=ActionStatus.succeeded,
+                        status=completed_status,
                         result_message=ctx._result_message,
                         result_metadata=ctx._result_metadata,
                     )
@@ -777,6 +805,10 @@ class Extractor(Generic[ConfigType], CogniteLogger):
                     result_message=str(e),
                 )
             )
+        finally:
+            with self._running_action_tokens_lock:
+                if self._running_action_tokens.get(action.external_id) is action_token:
+                    self._running_action_tokens.pop(action.external_id, None)
 
     def start(self) -> None:
         """
--- a/cognite/extractorutils/unstable/core/base.py
+++ b/cognite/extractorutils/unstable/core/base.py
@@ -781,11 +781,14 @@ class Extractor(Generic[ConfigType], CogniteLogger):
                     )
                 )
         except ActionError as e:
+            # As with start_task actions, a cooperative abort in response to cancellation may surface
+            # as an ActionError rather than a clean return; report that as canceled, not failed.
+            status = ActionStatus.canceled if action_token.is_cancelled else ActionStatus.failed
             filtered_metadata, oversized_fields = drop_oversized_metadata_fields(e.result_metadata)
             self._checkin_worker.queue_action_update(
                 ActionUpdate(
                     external_id=action.external_id,
-                    status=ActionStatus.failed,
+                    status=status,
                     result_message=(
                         str(e)
                         if not oversized_fields
@@ -798,10 +801,11 @@ class Extractor(Generic[ConfigType], CogniteLogger):
                 )
             )
         except Exception as e:
+            status = ActionStatus.canceled if action_token.is_cancelled else ActionStatus.failed
             self._checkin_worker.queue_action_update(
                 ActionUpdate(
                     external_id=action.external_id,
-                    status=ActionStatus.failed,
+                    status=status,
                     result_message=str(e),
                 )
             )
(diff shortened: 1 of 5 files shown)


cognite/extractorutils/unstable/core/base.py
```diff
@@ -203,6 +203,11 @@ def __init__(self, config: FullConfig[ConfigType], checkin_worker: CheckinWorker
         self._custom_actions: list[CustomAction] = []
         self._running_task_tokens: dict[str, CancellationToken] = {}
         self._running_task_tokens_lock = RLock()
+        # Keyed by Action.external_id (not task name) so a `cancel_pending` re-delivery of an
+        # in-flight start_task/custom action can be cancelled instead of re-dispatched. Populated by
+        # _handle_start_task_action/_handle_custom_action, consulted by _dispatch_single_action.
+        self._running_action_tokens: dict[str, CancellationToken] = {}
+        self._running_action_tokens_lock = RLock()
```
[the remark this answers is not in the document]
vikramlc-cognite: Thanks for flagging this - I don't think this race is actually reachable, though. **cancel_pending** for a given external_id is only ever produced by Odin after it has already recorded that action as running, which in turn requires Odin to have received our running ActionUpdate for it. 
In `_handle_start_task_action/_handle_custom_action`, the token is inserted into `_running_action_tokens` strictly before the running update is queued (same thread, sequential). So by the time Odin could possibly hand back a **cancel_pending** re-delivery for that id, the registration has already happened on our side — there's no window where the two orderings could invert. Given that, I'd rather not add **_cancelled_action_ids** for a race that can't occur under the current implementation.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -612,6 +617,16 @@ def _handle_actions(self, actions: list[Action]) -> None:
...
     def _dispatch_single_action(self, action: Action) -> None:
+        if action.status == ActionStatus.cancel_pending:
+            # Odin re-delivers an already-dispatched action once a user cancels it, with its status
+            # flipped to cancel_pending. This is not a fresh invocation: cancel the token tracking the
+            # in-flight run (if we have one) and stop, rather than re-running the action's handler.
+            with self._running_action_tokens_lock:
+                token = self._running_action_tokens.get(action.external_id)
+            if token is not None:
+                token.cancel()
+            return
```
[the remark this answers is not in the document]
vikramlc-cognite: Same [comment](https://github.com/cognitedata/python-extractor-utils/pull/572#discussion_r3860945755) as above

cognite/extractorutils/unstable/core/base.py
```diff
@@ -677,6 +695,10 @@ def _handle_start_task_action(self, action: Action) -> None:
                     result_message=str(e),
                 )
             )
+        finally:
```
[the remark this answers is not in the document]
vikramlc-cognite: Good catch on the general concern, but on closer inspection this specific path can't actually be reached the way described. **add_task** (`base.py:509-535`) wraps every task's real target in an internal **run_task** function that already catches all exceptions from the task itself, reports them separately via the fatal-error pipeline (task_context.exception(...)), and never re-raises. So by the time _run_task_with_token returns, a task that raised due to cancellation has already had its exception fully absorbed - **_run_task_with_token** returns normally, and the existing success-path check `(status = ActionStatus.canceled if child_token.is_cancelled else ActionStatus.succeeded`, just above this except block) already reports canceled correctly today.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -777,6 +805,10 @@ def _handle_custom_action(self, action: Action) -> None:
...
```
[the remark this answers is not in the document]
vikramlc-cognite: Genuine. Fixed.

cognite/extractorutils/unstable/core/base.py
jagadish-bihani: why is this ActionStatus failed here, shouldn't it be completed_status?

vikramlc-cognite: Good catch, you are right, this should be **completed_status**, not a hardcoded failed. Updated it and also customised the **result_message** to reflect cancelled or completed.

jagadish-bihani (dismissed):

cognite/extractorutils/unstable/core/base.py
```diff
@@ -720,23 +746,28 @@ def _handle_custom_action(self, action: Action) -> None:
...
+                outcome = "was canceled" if completed_status == ActionStatus.canceled else "completed successfully"
                 self._checkin_worker.queue_action_update(
                     ActionUpdate(
                         external_id=action.external_id,
-                        status=ActionStatus.failed,
+                        status=completed_status,
                         result_message=truncate_message(
-                            f"Action '{custom.name}' completed successfully, but metadata field(s) "
+                            f"Action '{custom.name}' {outcome}, but metadata field(s) "
                             f"{', '.join(oversized_fields)} exceeded the {MAX_METADATA_VALUE_BYTES}-byte-per-value "
```
Jacob-Eliat-Eliat, line 788: (out of review scope) How long can the oversized field array be? Might make sense to truncate it to something reasonable so that you don't miss the end of the error message due to truncate_message which would make the error message impossible to understand.

Of course this is completely out of scope of the review (so not a blocker for merge), just something I noticed.

vikramlc-cognite: Agreed this is out of scope here but I'll track it as a follow-up PR (reorder the message so the fixed explanation comes before the unbounded field list, so truncation only ever eats into the list of names rather than the reason itself, and probably report the count up front). Thanks for flagging it.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -661,6 +676,9 @@ def _handle_start_task_action(self, action: Action) -> None:
                 return
             self._running_task_tokens[task_name] = child_token
 
+        with self._running_action_tokens_lock:
+            self._running_action_tokens[action.external_id] = child_token
```
Jacob-Eliat-Eliat: This might be very theoretical since there's nothing likely to throw in between, but ideally the assignments to _running_task_tokens and   _running_action_tokens should happen inside the same try that cleans them up in finally.

Right now the only thing preventing a stuck task/action entry is that nothing between registration and the try currently throws which is not very future proof.

_It's in my opinion a bit debatable: one one hand best practice is to have assignment to the map inside the try-finally because if finally is not run then there is a mem-leak, on the other hand, it's bad practice to wrap too much non-throwing code in a try block. Looking forwards to mentor review on this._

vikramlc-cognite: Agreed, I traced through it and this is worth fixing, for a slightly sharper reason than the leaked-entry framing: if something in that gap does throw, it doesn't just leak a dict entry, it leaves the task permanently stuck **running** until the whole extractor restarts.

cognite/extractorutils/unstable/core/base.py
```diff
@@ -711,32 +749,42 @@ def _handle_custom_action(self, action: Action) -> None:
...
+                # (which would otherwise poison the checkin batch and retry forever, since checkin
+                # bundles all pending updates together and requeues the whole batch on any rejection).
+                # Non-oversized fields are still reported, and the action's real outcome (succeeded or
+                # canceled) is preserved rather than being overwritten by this reporting issue.
+                outcome = "was canceled" if completed_status == ActionStatus.canceled else "completed successfully"
                 self._checkin_worker.queue_action_update(
                     ActionUpdate(
                         external_id=action.external_id,
-                        status=ActionStatus.failed,
+                        status=completed_status,
```
Jacob-Eliat-Eliat, line 785: paraphrasing in advance what I think @dmivankov will say in his mentor review (from briefly discussing with him):

the completed status changed from failed to succeeded, which according to context looks fine, but these kind of changes should really be in the PR's description.

Not blocking the PR, but as a reviewer it feels like worth flagging that there is a major behavioral change here.

Jacob-Eliat-Eliat:
🐴 next step:  will need to do a mentor review, flagging anything that I could have missed before it can be merged.

cognite/extractorutils/unstable/core/base.py
vikramlc-cognite: Updated

nithinb (approved):
:unicorn: LGTM

